PR #1, opened 2021-12-21
Validate the public keys that are used for the SAS shared secret computation

poljar:


src/sas.rs
```diff
@@ -73,6 +76,16 @@ pub enum SasError {
     Mac(#[from] MacError),
 }
 
+/// Error type describing failures that can happen when we try to create a
+/// shared secret.
+#[derive(Debug, Error)]
+pub enum PublicKeyError {
+    #[error(transparent)]
+    PublicKey(#[from] Curve25519KeyError),
+    #[error("One of the keys did not have contributory behaviour")]
```
dkasak, line 85:
```suggestion
    #[error("At least one of the keys did not have contributory behaviour")]
```

src/sas.rs
```diff
@@ -200,23 +213,33 @@ impl Sas {
...
+        } else {
+            Err(PublicKeyError::NonContributoryKey)
+        }
     }
 
     /// Establishes a SAS secret by performing a DH handshake with another
     /// public key in "raw", base64-encoded form.
+    ///
+    /// Returns an [`EstablishedSas`] object which can be used to generate
+    /// [`SasBytes`] if the given public key was valid, otherwise `None`.
```
dkasak, line 236:
```suggestion
    /// [`SasBytes`] if the received public key is valid, otherwise `None`.
```

dkasak (approved):

PR #6, opened 2021-12-28
Use a ReusableSecret for the Sas curve25519 secret key

poljar:
Methods that take self can't easily be bound over FFI, since the
intetion of this crate is to work in various languages let's switch to a
method that takes a borrow of self.

This is a bit unfortunate since we do not want people to reuse the
secret key here.

### DIFF
--- a/src/sas.rs
+++ b/src/sas.rs
@@ -222,7 +222,7 @@ impl Sas {
     ///
     /// # Security
     ///
-    /// Please note that this method *must* only be called once, if the method
+    /// Please note that this method *must* be called only once. If the method
     /// fails or we receive another public key another [`Sas`] object needs to
     /// be created.
     pub fn diffie_hellman(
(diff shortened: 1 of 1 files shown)


src/sas.rs
```diff
@@ -219,8 +219,14 @@ impl Sas {
     ///
     /// Returns an [`EstablishedSas`] object which can be used to generate
     /// [`SasBytes`] if the given public key was valid, otherwise `None`.
+    ///
+    /// # Security
+    ///
+    /// Please note that this method *must* only be called once, if the method
+    /// fails or we receive another public key another [`Sas`] object needs to
+    /// be created.
```
dkasak, line 227:
```suggestion
    /// Please note that this method *must* be called only once. If the method
    /// fails or we receive another public key another [`Sas`] object needs to
    /// be created.
```

src/sas.rs
```diff
@@ -237,8 +243,13 @@ impl Sas {
     ///
     /// Returns an [`EstablishedSas`] object which can be used to generate
     /// [`SasBytes`] if the received public key is valid, otherwise `None`.
+    ///
+    /// # Security
+    ///
+    /// Please note that this method *must* only be called once, if the method
+    /// fails or we receive another public key another [`Sas`] object needs to
```
dkasak, line 250:
```suggestion
    /// Please note that this method *must* be called only once. If the method
    /// fails or we receive another public key another [`Sas`] object needs to
    /// be created.
```

PR #15, opened 2022-01-06
Add a feature to enable strict signature verification

poljar:
This requires #14.


poljar:
This has been merged, I forgot to change the base branch so github doesn't notice it.

PR #18, opened 2022-01-07
Simplify our serialization logic

poljar:
A lot of types in our serialization logic seem to be of questionable usage. We should revise our serialization logic and where possible remove helper structs and use serde directly, this include error types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,8 @@ only *unpickling* is supported.
 
 ### Modern pickles
 
-The modern pickling mechanism used by this crate. The exact serialization
-format which will be used is undecided but for now we're pickling to JSON.
-Since the pickling support is based on `serde`, changing the format is easy.
+The modern pickling mechanism used by this crate. The serialization format is
+based on Serde.
 
 The following structs support pickling:
 
@@ -131,18 +130,14 @@ The following structs support pickling:
 - `GroupSession`
 - `InboundGroupSession`
 
-To pickle into a JSON string, simply call the `.pickle_to_json_string()` method,
-which will return a special struct implementing `.as_str()`,
-`Deref<Target=str>` and `AsRef<str>` which you can use to get to the actual
-serialized string. This struct will zeroize its memory once it drops so that
-any secrets within do not linger on.
-
 For example, the following will print out the JSON representing the serialized
 `Account` and will leave no new copies of the account's secrets in memory:
 
 ```rust
 use anyhow::Result;
-use vodozemac::olm::Account;
+use vodozemac::olm::{Account, AccountPickle};
+
+const PICKLE_KEY: [u8; 32] = [0u8; 32];
 
 fn main() -> Result<()>{
     let mut account = Account::new();
@@ -150,11 +145,11 @@ fn main() -> Result<()>{
     account.generate_one_time_keys(10);
     account.generate_fallback_key();
 
-    let pickle = account.pickle_to_json_string();
+    let pickle = account.pickle().encrypt(&PICKLE_KEY);
 
-    print!("{}", pickle.as_str());
+    let account2: Account = AccountPickle::from_encrypted(&pickle, &PICKLE_KEY)?.into();
 
-    let account2 = pickle.unpickle()?;  // this is the same as `account`
+    assert_eq!(account.identity_keys(), account2.identity_keys());
 
     Ok(())
 }
@@ -164,12 +159,13 @@ You can unpickle a pickle-able struct directly from its serialized form:
 
 ```rust
 # use anyhow::Result;
-# use vodozemac::olm::Account;
+# use vodozemac::olm::{Account, AccountPickle};
 #
 # fn main() -> Result<()> {
 #   let some_account = Account::new();
-    let json_str = some_account.pickle_to_json_string();
-    let account: Account = serde_json::from_str(&json_str)?;  // the same as `some_account`
+    let json_str = serde_json::to_string(&some_account.pickle())?;
+    // This will produce an account which is identitcal to `some_account`.
+    let account: Account = serde_json::from_str::<AccountPickle>(&json_str)?.into();
 #
 #    Ok(())
 # }
@@ -188,7 +184,7 @@ use vodozemac::olm::Account;
 
 fn main() -> Result<()> {
     let account = Account::new();
-    let account: Account = account.pickle().unpickle()?;  // this is identity
+    let account: Account = account.pickle().into();  // this is identity
 
     Ok(())
 }
--- a/README.md
+++ b/README.md
@@ -120,8 +120,10 @@ only *unpickling* is supported.
 
 ### Modern pickles
 
-The modern pickling mechanism used by this crate. The serialization format is
-based on Serde.
+The crate also implements a modern pickling mechanism using
+[Serde](https://serde.rs/). The exact serialization format is not mandated or
+specified by this crate, but you can serialize to and deserialize from any
+format supported by Serde.
 
 The following structs support pickling:
 
@@ -160,12 +162,15 @@ You can unpickle a pickle-able struct directly from its serialized form:
 ```rust
 # use anyhow::Result;
 # use vodozemac::olm::{Account, AccountPickle};
+# use zeroize::Zeroize;
 #
 # fn main() -> Result<()> {
 #   let some_account = Account::new();
-    let json_str = serde_json::to_string(&some_account.pickle())?;
-    // This will produce an account which is identitcal to `some_account`.
+    let mut json_str = serde_json::to_string(&some_account.pickle())?;
+    // This will produce an account which is identical to `some_account`.
     let account: Account = serde_json::from_str::<AccountPickle>(&json_str)?.into();
+
+    json_str.zeroize();
 #
 #    Ok(())
 # }
--- a/src/megolm/group_session.rs
+++ b/src/megolm/group_session.rs
@@ -20,7 +20,7 @@ use crate::{
     cipher::Cipher,
     types::Ed25519Keypair,
     utilities::{base64_encode, pickle, unpickle},
-    UnpickleError,
+    PickleError,
 };
 
 /// A Megolm group session represents a single sending participant in an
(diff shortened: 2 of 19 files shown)


src/megolm/ratchet.rs
```diff
@@ -199,38 +245,10 @@ impl Ratchet {
...
-            counter: pickle.counter,
-        })
-    }
-}
-
 #[derive(Error, Debug)]
-pub enum MegolmRatchetUnpicklingError {
-    #[error("Invalid Megolm ratchet key length: expected 128, got {0}")]
-    InvalidKeyLength(usize),
+pub enum RatchetError {
```
dkasak, line 249: Do we expect to ever use this outside deserialization? If not, can we please name it something like `RatchetDeserializationError`?

poljar: But it's not a deserialization error. It is used in the deserialization implementation and converted into a `serde::de::Error::custom`. I don't think this makes much sense and the type should anyways be private to this module.

src/megolm/group_session.rs
```diff
@@ -131,86 +127,34 @@ impl GroupSession {
...
-    type Error = GroupSessionUnpicklingError;
-
-    fn try_from(pickle: GroupSessionPickle) -> Result<Self, Self::Error> {
-        Ok(Self { ratchet: pickle.ratchet.try_into()?, signing_key: pickle.signing_key })
-    }
-}
-
-#[derive(Zeroize, Debug)]
-#[zeroize(drop)]
-pub struct GroupSessionPickledJSON(String);
```
dkasak, line 181: The point of this type was to avoid people serializing to `String` which would then be left unzeroized since `String` doesn't implement zeroization on drop. Instead, this wrapper type offered to handle this for you while providing a way to access the raw serialized content via borrowing.

However, from internal discussions with @poljar, we're worried about the proliferation of types being hard to explain and hard to use. Instead, we're considering placing prominent warnings in the documentation that users of pickling functionality should take care to zeroize their serialization buffers themselves or use the `encrypt` method.

Can we please add a TODO to add such docs?

src/olm/account/mod.rs
```diff
@@ -527,70 +516,32 @@ pub struct AccountPickle {
...
 }
 
 /// A format suitable for serialization which implements [`serde::Serialize`]
 /// and [`serde::Deserialize`]. Obtainable by calling [`Account::pickle`].
 impl AccountPickle {
-    /// Convert the pickle format back into an [`Account`].
-    pub fn unpickle(self) -> Result<Account, AccountUnpicklingError> {
-        self.try_into()
+    pub fn encrypt(self, pickle_key: &[u8; 32]) -> String {
+        pickle(&self, pickle_key)
```
dkasak, line 526: Missing docs.

src/megolm/inbound_group_session.rs
```diff
@@ -344,64 +329,36 @@ impl InboundGroupSession {
...
-                .map_err(InboundGroupSessionUnpicklingError::InvalidInitialRatchet)?,
-            latest_ratchet: pickle
-                .latest_ratchet
-                .try_into()
-                .map_err(InboundGroupSessionUnpicklingError::InvalidLatestRatchet)?,
-            signing_key: pickle.signing_key,
-            signing_key_verified: pickle.signing_key_verified,
-        })
+    pub fn from_encrypted(ciphertext: &str, pickle_key: &[u8; 32]) -> Result<Self, PickleError> {
+        unpickle(ciphertext, pickle_key)
```
dkasak, line 344: The meaning of the method should be pretty clear, but let's still add a doc comment for the method.

src/megolm/inbound_group_session.rs
```diff
@@ -344,64 +329,36 @@ impl InboundGroupSession {
...
     signing_key: Ed25519PublicKey,
     #[allow(dead_code)]
     signing_key_verified: bool,
 }
 
 impl InboundGroupSessionPickle {
-    pub fn unpickle(self) -> Result<InboundGroupSession, InboundGroupSessionUnpicklingError> {
-        self.try_into()
+    pub fn encrypt(self, pickle_key: &[u8; 32]) -> String {
+        pickle(&self, pickle_key)
```
dkasak, line 340: Docs.

README.md
```diff
@@ -164,12 +159,13 @@ You can unpickle a pickle-able struct directly from its serialized form:
...
-# use vodozemac::olm::Account;
+# use vodozemac::olm::{Account, AccountPickle};
 #
 # fn main() -> Result<()> {
 #   let some_account = Account::new();
-    let json_str = some_account.pickle_to_json_string();
-    let account: Account = serde_json::from_str(&json_str)?;  // the same as `some_account`
+    let json_str = serde_json::to_string(&some_account.pickle())?;
+    // This will produce an account which is identitcal to `some_account`.
+    let account: Account = serde_json::from_str::<AccountPickle>(&json_str)?.into();
```
dkasak, line 168: Just so it's clearer that the caller has to zeroize:

```suggestion
    let account: Account = serde_json::from_str::<AccountPickle>(&json_str)?.into();
    
    json_str.zeroize();
```

src/types/curve25519.rs
```diff
@@ -156,33 +153,22 @@ impl<'a> From<&'a ReusableSecret> for Curve25519PublicKey {
...
-    public: [u8; Curve25519PublicKey::LENGTH],
-}
-
-impl Drop for Curve25519KeypairPickle {
-    fn drop(&mut self) {
-        self.secret.zeroize();
-    }
-}
+#[serde(transparent)]
+pub(crate) struct Curve25519KeypairPickle(Curve25519SecretKey);
```
dkasak, line 157: Sadly, I think we have to box this so that zeroization occurs correctly since `Curve25519SecretKey` is just a newtype-style wrapper for an array. If it's not boxed, then moving the pickle can make a copy of the array which won't get properly zeroized.

dkasak (changes requested):
Left a bunch of comments. In the interest of getting this out as soon as possible, if any of my suggestions here are hard to do right now, I think the only critical thing is the boxing of `Curve25519SecretKey`. The rest can be done in a separate clean up PR if need be.

src/megolm/inbound_group_session.rs
```diff
@@ -344,64 +329,36 @@ impl InboundGroupSession {
...
-impl Deref for InboundGroupSessionPickledJSON {
-    type Target = str;
-
-    fn deref(&self) -> &Self::Target {
-        self.as_ref()
+impl From<InboundGroupSessionPickle> for InboundGroupSession {
+    fn from(pickle: InboundGroupSessionPickle) -> Self {
+        Self {
+            initial_ratchet: pickle.initial_ratchet.clone(),
+            latest_ratchet: pickle.initial_ratchet,
```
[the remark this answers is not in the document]
poljar: `latest_ratchet` gets reset to the `initial_ratchet` any time the `latest_ratchet` is advanced past the message index of the message that is being decrypted.

I don't think `latest_ratchet` plays part in the identity of the `InboundGroupSession`, we don't need to be byte for byte the same after a roundtrip.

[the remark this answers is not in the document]
dkasak: I'm aware of that, but we do have logic dependent on values of `initial_ratchet` and `latest_ratchet`. Introducing changes to `latest_ratchet` during a serialization cycle then complicates reasoning and introduces opportunities for unforeseen effects on that logic.

I guess I'm asking for a rationale on why we'd not want to save the state and remove this factor completely. Is it purely to make the pickle smaller or something else?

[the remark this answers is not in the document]
poljar: > I'm aware of that, but we do have logic dependent on values of `initial_ratchet` and `latest_ratchet`. Introducing changes to `latest_ratchet` during a serialization cycle then complicates reasoning and introduces opportunities for unforeseen effects on that logic.

Could you clarify what you're talking about here? What kind of logic?

> I guess I'm asking for a rationale on why we'd not want to save the state and remove this factor completely. Is it purely to make the pickle smaller or something else?

Yes, we're wasting 128 bytes per inbound group session otherwise. I know that libolm pickles the memory as is and wastes a lot space that way. It's not something I'm keen on porting.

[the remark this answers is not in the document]
dkasak: > Could you clarify what you're talking about here? What kind of logic?

Sorry, I forgot to link. I was talking about e.g. `find_ratchet` (https://github.com/matrix-org/vodozemac/blob/bde1fbc96557d3b77ec1a6f92696468d155d2a4c/src/megolm/inbound_group_session.rs#L187) and `advance_to`.

> Yes, we're wasting 128 bytes per inbound group session otherwise. I know that libolm pickles the memory as is and wastes a lot space that way. It's not something I'm keen on porting.

I guess that's fair, though this way we need to recalculate things we already calculated instead, though that may be less of a problem. It's just something to be mindful of if we do decide to keep this; I don't feel too strongly either way.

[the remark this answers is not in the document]
poljar: This is the code I was talking about as well, `latest_ratchet` will regularly be reset to `initial_ratchet`, so I don't think that there's anything to worry about. It is also set to `initial_ratchet` when we first create the the session.

In my view, `latest_ratchet` is only an optimization so that we don't need to advance from the start every time we decrypt a message, from the megolm spec:

>  They may also choose to cache calculated ratchet values,
but the decision of which ratchet states to cache is left to the application.

src/megolm/group_session.rs
poljar: Opened an issue https://github.com/matrix-org/vodozemac/issues/45

src/types/curve25519.rs
poljar: There seem to be a couple of such problems, so this will be handled in a separate PR.

src/megolm/ratchet.rs
poljar: Did so.

dkasak (approved):
Looks good now. Pushed a commit tweaking doc comments.

PR #19, opened 2022-01-07
Add repository to Cargo.toml

erikjohnston:
So that crates.io includes a link to the repo.

@poljar: should I be requesting review from you or is the a team I should be using?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ authors = [
 ]
 name = "vodozemac"
 description = "A Rust implementation of Olm and Megolm"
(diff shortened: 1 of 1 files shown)


poljar:
> So that crates.io includes a link to the repo.
> 
> @poljar: should I be requesting review from you or is the a team I should be using?

We don't have a formal team for this currently so it's fine to request from me. I'll be away next week until the end of the month so if you have more PRs comming @dkasak would probably be the person to request from.

poljar (approved):

PR #22, opened 2022-01-12
Fix OTK pickling

erikjohnston:
We remove the OTKs from the `public_keys` map once they're published, but we need to keep the entries in `reverse_public_keys` for as long as they're still in `private_keys`.

Not sure if we want to rename `reverse_public_keys`?


dkasak:
Ah, thanks for spotting this. The change looks good, but we probably want to either add a comment explaining this or go through with the renaming.

> Not sure if we want to rename `reverse_public_keys`?

Sure, any ideas for the name? Does `key_ids_by_key` sound intelligible to you?

dkasak (approved):
I'll handle the rename separately.

erikjohnston:
> I'll handle the rename separately.

Oh, I just did that in #24 :)

PR #28, opened 2022-01-25
vodozemac: Switch to bincode to decode libolm pickles

poljar:
This PR switches towards using the bincode crate to decode libolm pickles. This requires a bincode patch, since bincode encodes slice lengths as u64 while libolm as u32.


poljar:
PR to add u32 slice lengths to bincode: https://github.com/bincode-org/bincode/pull/493

poljar:
Bincode support isn't happening, closing this.

PR #30, opened 2022-01-31
feat(megolm): Allow inbound group sessions to be advanced permanently

### LINKED ISSUE #26: Add ability to permanently advance ratchet of Megolm inbound group session
We should add a way to permanently advance the ratchet of an inbound group session. Currently we keep the initial ratchet state forever.

It's possible to work around this limitation currently by using a pair of `InboundGroupSession::export_at` and `InboundGroupSession::import` calls in order to produce a ratcheted copy of the session, but we should also support doing this in-place.

poljar:
This closes #26.


dkasak (approved):
Looks good.

PR #33, opened 2022-02-09
Sas libolm compat

poljar:
This PR adds a dedicated MAC type to libolm, useful because Ruma nowadays decodes base64 encoded strings when deserializing. 

It also adds the buggy mac calculation method from libolm, since that one is still widely used.


dkasak (approved):
Looks good. I left some comment nits.

PR #35, opened 2022-02-14
Add and expose an ed25519 secret key type

poljar:
This type is useful if users need to create a separate identity to sign messages with, i.e. it's used in cross signing in Matrix land.

### DIFF
--- a/src/types/ed25519.rs
+++ b/src/types/ed25519.rs
@@ -37,8 +37,8 @@ pub enum SignatureError {
 #[derive(Deserialize, Serialize)]
 #[serde(try_from = "Ed25519KeypairPickle")]
 #[serde(into = "Ed25519KeypairPickle")]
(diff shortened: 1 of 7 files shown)


src/types/ed25519.rs
```diff
@@ -57,48 +59,130 @@ impl Ed25519Keypair {
...
+
+impl Default for Ed25519SecretKey {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
 #[derive(Serialize, Deserialize)]
-enum Ed25519SecretKey {
+enum SecretKeys {
```
dkasak, line 166: Why the plural here?

poljar: It avoids a name clash with `ed25519_dalek::SecretKey`.

Feel free to change it if you feel that strongly about it.

PR #43, opened 2022-02-22
Fix zeroization issues.

dkasak:


src/olm/session/mod.rs
```diff
@@ -437,12 +439,15 @@ impl Session {
                 }
 
                 if let Some(chain) = pickle.sender_chains.get(0) {
+                    // XXX: Passing in secret array as value.
```
poljar, line 442: Can we make these a bit more descriptive. If someone else comes along, they won't have a clue what those are about.

src/olm/account/one_time_keys.rs
```diff
@@ -32,6 +32,14 @@ pub(super) struct OneTimeKeys {
     pub key_ids_by_key: HashMap<Curve25519PublicKey, KeyId>,
 }
 
+impl Drop for OneTimeKeys {
```
poljar, line 35: Is this needed? All the values in the map will get dropped and zeroize is implemented for the type in the map.

src/megolm/ratchet.rs
```diff
@@ -208,22 +208,21 @@ pub(super) struct RatchetPickle {
...
-        Self { key: ratchet.inner.into(), counter: ratchet.counter }
+        Self { key: (*ratchet.inner).into(), counter: ratchet.counter }
     }
 }
 
 impl TryFrom<RatchetPickle> for Ratchet {
     type Error = MegolmRatchetUnpicklingError;
 
     fn try_from(pickle: RatchetPickle) -> Result<Self, Self::Error> {
-        Ok(Ratchet {
```
poljar, line 219: Need to remember that we need to take another look at this after the simplified pickling branch is merged.

src/olm/session/mod.rs
```diff
@@ -327,8 +327,8 @@ impl Session {
         #[derive(Debug, Zeroize)]
         #[zeroize(drop)]
         struct ReceivingChain {
```
poljar, line 329: Didn't you intend to rename those so it's clearer they are libolm specific?

poljar (changes requested):
Looks good, left a couple of nits.

src/olm/session/mod.rs
dkasak: I think there may be more renames necessary for clarity so I decided not to bother with it in this PR and leave it for another one.

src/olm/account/one_time_keys.rs
dkasak: Yep, thanks for catching. What I actually should've done is remove the `Zeroize` impl for `OneTimeKeysPickle` to make everything symmetric. I'll make the change.

src/megolm/ratchet.rs
dkasak: Added a TODO.

poljar (approved):

PR #49, opened 2022-03-09
Expose some lower level APIs behind a `low-level` feature flag.

erikjohnston:
This is so that advanced systems can do non-standard stuff, without exposing those APIs by standard.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,9 @@ default = ["libolm-compat"]
 js = ["rand/wasm-bindgen"]
 strict-signatures = []
 libolm-compat = []
+# The low-level-api feature exposes extra APIs that are only useful in advanced
+# use cases and require extra care to use.
+low-level-api = []
 
 [dependencies]
 aes = "0.7.5"
--- a/src/hazmat/mod.rs
+++ b/src/hazmat/mod.rs
@@ -1,4 +1,4 @@
-// Copyright 2021 The Matrix.org Foundation C.I.C.
+// Copyright 2022 The Matrix.org Foundation C.I.C.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/hazmat/mod.rs
+++ b/src/hazmat/mod.rs
@@ -1,5 +1,4 @@
 // Copyright 2021 The Matrix.org Foundation C.I.C.
-// Copyright 2021 Damir Jelić, Denis Kasak
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! An implementation of the Olm ratchet.
+//! ⚠️ Low-level "hazmat" functions.
+//!
+//! This module contains low level APIs that should *not* be used or needed by
+//! most users.
+//!
+//! These functions are exported to aid very advanced use cases.
 
-mod account;
-mod messages;
-pub(crate) mod session;
-mod session_keys;
-mod shared_secret;
+#![cfg(feature = "low-level-api")]
 
-pub use account::{Account, AccountPickle, IdentityKeys, InboundCreationResult};
-pub use messages::{Message, MessageType, OlmMessage, PreKeyMessage};
-pub use session::{ratchet::RatchetPublicKey, DecryptionError, Session, SessionPickle};
-pub use session_keys::SessionKeys;
+pub mod olm;
+
+pub use crate::cipher::{Cipher, Mac};
--- a/src/megolm/inbound_group_session.rs
+++ b/src/megolm/inbound_group_session.rs
@@ -174,9 +174,9 @@ impl InboundGroupSession {
         }
     }
 
-    #[cfg(feature = "low-level-api")]
     /// Returns a copy of the [`Cipher`] at the given message index, without
     /// advancing the internal ratchets.
+    #[cfg(feature = "low-level-api")]
     pub fn get_cipher_at(&self, message_index: u32) -> Option<Cipher> {
         if self.initial_ratchet.index() >= message_index {
             let mut ratchet = self.initial_ratchet.clone();
--- a/src/megolm/inbound_group_session.rs
+++ b/src/megolm/inbound_group_session.rs
@@ -174,6 +174,19 @@ impl InboundGroupSession {
         }
     }
 
+    #[cfg(feature = "low-level-api")]
+    pub fn get_cipher_at(&self, message_index: u32) -> Option<Cipher> {
+        if self.initial_ratchet.index() >= message_index {
+            let mut ratchet = self.initial_ratchet.clone();
+            if self.initial_ratchet.index() > message_index {
+                ratchet.advance_to(message_index);
+            }
+            Some(Cipher::new_megolm(ratchet.as_bytes()))
+        } else {
+            None
+        }
+    }
+
     fn find_ratchet(&mut self, message_index: u32) -> Option<&Ratchet> {
         if self.initial_ratchet.index() == message_index {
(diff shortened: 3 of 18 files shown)


Cargo.toml
```diff
@@ -24,6 +24,9 @@ default = ["libolm-compat"]
 js = ["rand/wasm-bindgen"]
 strict-signatures = []
 libolm-compat = []
+# The low-level-api feature exposes extra APIs that are only useful in advanced
```
poljar, line 27: This need to be documented in the readme.

src/megolm/inbound_group_session.rs
```diff
@@ -174,6 +174,21 @@ impl InboundGroupSession {
         }
     }
 
+    #[cfg(feature = "low-level-api")]
```
poljar, line 177: Please put the cfg attributes bellow the docs.

src/megolm/message.rs
```diff
@@ -113,6 +140,25 @@ impl MegolmMessage {
         message.encode_manual()
     }
 
+    /// Create a new [`MegolmMessage`] with the given plaintext and keys.
+    pub fn encrypt(
```
poljar, line 144: This needs to be put behind the feature flag as well, no?

src/megolm/message.rs
```diff
@@ -104,6 +116,21 @@ impl MegolmMessage {
         base64_encode(self.to_bytes())
     }
 
+    #[cfg(feature = "low-level-api")]
```
poljar, line 119: Bellow the docs please.

src/olm/messages/pre_key.rs
```diff
@@ -132,13 +133,19 @@ impl PreKeyMessage {
...
     }
 
-    pub(crate) fn new(
-        one_time_key: Curve25519PublicKey,
-        base_key: Curve25519PublicKey,
-        identity_key: Curve25519PublicKey,
-        message: Message,
-    ) -> Self {
-        Self { one_time_key, base_key, identity_key, message }
+    #[cfg(feature = "low-level-api")]
```
poljar, line 136: Again.

src/olm/session/message_key.rs
```diff
@@ -43,22 +43,43 @@ impl Drop for RemoteMessageKey {
 }
 
 impl MessageKey {
+    #[cfg(feature = "low-level-api")]
```
poljar, line 46: Do we need to change the constructor? Can't we:

1. Make the `MessageKey` pub
2. Only re-export it publicly if the feature flag is set.

src/olm/session/mod.rs
```diff
@@ -207,17 +209,25 @@ impl Session {
...
             OlmMessage::PreKey(message)
         }
     }
 
+    /// Get the keys associated with this session.
+    pub fn session_keys(&self) -> SessionKeys {
+        self.session_keys
+    }
+
+    #[cfg(feature = "low-level-api")]
```
poljar, line 223: Bellow the docs.

src/olm/session/mod.rs
```diff
@@ -207,17 +209,25 @@ impl Session {
...
+    /// Get the [`MessageKey`] to encrypt the next message.
+    ///
+    /// **Note**: This *must* be used to encrypt the next message.
```
poljar, line 226: This isn't strictly true, a `Session` supports out of order decryption, you just need to make sure you're not too much out of order. Though not sure if we want to signal that to the user.

src/olm/session/message_key.rs
```diff
@@ -43,22 +43,43 @@ impl Drop for RemoteMessageKey {
...
 
         message
     }
+
+    /// Get a reference to the message key's key.
+    pub fn key(&self) -> &[u8; 32] {
+        self.key.as_ref()
+    }
+
+    /// Get a reference to the message key's ratchet key.
```
poljar, line 74: `RatchetPublicKey` is now `Copy`, do we still want to return a reference?

poljar (changes requested):
Looks good, couple of nits and perhaps we should collect the dangerous types that otherwise aren't public into a separate module.

Something like what the rust crypto crates are doing, i.e. for the [aes](https://docs.rs/aes/latest/aes/hazmat/index.html) crate.

Create a `hazmat` module and re-export the dangerous types only from there. That way it's clearer which types shouldn't be used, even though the feature flag made that already clear.

It also cleans up the main page from types normal users don't need to concern themselves with.

src/olm/session/mod.rs
erikjohnston: What about "This must be used to encrypt a message"?

src/hazmat/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+//! ⚠️ Low-level "hazmat" functions.
+//!
+//! This module contains low level APIs that should *not* be used or needed by
+//! most users.
+//!
+//! These functions are exported to aid very advanced use cases.
+
+#![cfg(feature = "low-level-api")]
+
+pub mod olm;
```
poljar, line 24: I wonder if this is a bit much considering that there's a single type in the olm module. You plan on having a megolm variant of the message key as well? Not really important to change this if you don't feel like it.

src/olm/session/mod.rs
```diff
@@ -207,17 +212,25 @@ impl Session {
...
     }
 
+    /// Get the keys associated with this session.
+    pub fn session_keys(&self) -> SessionKeys {
+        self.session_keys
+    }
+
+    /// Get the [`MessageKey`] to encrypt the next message.
+    ///
+    /// **Note**: This *must* be used to encrypt the message.
```
poljar, line 228: Perhaps something like:

> **Note**: This *should* be used to encrypt the next message, `Session`s support only a limited amount of out of order messages before decryption starts to fail.

poljar (changes requested):
Almost there, please wake up and remember what year it is, it's already March, you can't keep on doing this to yourself :stuck_out_tongue_closed_eyes:.

I also proposed a doc comment for `next_message_key()` that hopefully makes both of us happy.

src/olm/session/mod.rs
dkasak: I *think* the intent is to express this, so how about something a bit more detailed?

```suggestion
    /// **Note**: Each key obtained in this way should be used to encrypt
    /// a message and the message must then be sent to the recipient.
    ///
    /// Failing to do so will increase the number of out-of-order messages on
    /// the recipient side. Given that a `Session` can only support a limited
    /// number of out-of-order messages, this will eventually lead to
    /// undecryptable messages.
```

src/hazmat/mod.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright 2021 The Matrix.org Foundation C.I.C.
```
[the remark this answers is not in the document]
erikjohnston: Look, I *intended* to update the year!  That must count for something! :stuck_out_tongue:

src/hazmat/mod.rs
erikjohnston: I'm persuadable, but my main thought was that a lot of the types in `olm` aren't obviously named in an olm specific way, e.g. `MessageKey` could equally by related to megolm or whatever. :shrug:

src/olm/session/mod.rs
erikjohnston: Love it, thank you

erikjohnston: Hilariously github has attributed that suggestion to @poljar.... https://github.com/matrix-org/vodozemac/pull/49/commits/0f1d1e530f6f199bb4ba884aa0be0c5bc21aa429

src/hazmat/mod.rs
poljar: Yeah, let's leave it.

src/olm/session/mod.rs
poljar: As it should, thanks github.

poljar (approved):
Thanks, looks good.

PR #52, opened 2022-04-01
Fix bug in `get_cipher_at`

erikjohnston:
I got the operator the wrong way round :(

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,4 +97,5 @@ jobs:
       - name: Test
         uses: actions-rs/cargo@v1
         with:
-          command: test --all-features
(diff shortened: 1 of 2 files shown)


poljar:
Could we get a test for this? I would really like to have tests for every bugfix we push to vodozemac. I know this is trivial but it would still be nice to have a test in case we ever start refactoring things.

erikjohnston:
> Could we get a test for this? I would really like to have tests for every bugfix we push to vodozemac. I know this is trivial but it would still be nice to have a test in case we ever start refactoring things.

Quite right, have done this now

poljar (approved):
Looks good, thanks.

PR #54, opened 2022-04-06
Use the SessionKeys struct in pre-key messages

poljar:


src/olm/messages/pre_key.rs
```diff
@@ -44,23 +42,29 @@ impl PreKeyMessage {
...
     /// to establish a [`Session`]
     ///
     /// [`Session`]: crate::olm::Session
     pub fn identity_key(&self) -> Curve25519PublicKey {
-        self.identity_key
+        self.session_keys.identity_key
+    }
+
+    /// The keys that can be used to establish an Olm [`Session`] from this
+    /// pre-key message.
```
dkasak, line 65:
```suggestion
    /// The collection of all keys required for establishing an Olm [`Session`]
    /// from this pre-key message.
    ///
    /// Other methods on this struct (like [`identity_key`]) can be used to
    /// retrieve individual keys from this collection.
    ///
    /// [`Session`]: crate::olm::Session
```

dkasak (approved):
One suggestion but otherwise good to go.

PR #62, opened 2022-05-17
Remove encoded public key variants of some methods

poljar:
- refactor(types): Remove the encoded public key from the curve25519 types
- refactor(sas): Remove the encoded public key methods
- refactor(olm): Remove the encoded one-time keys method


dkasak (approved):
LGTM

PR #72, opened 2022-06-22
feat: Expose encrypting and decrypting bytes

zaynetro:
This is a proposal to allow using bytes directly without extra
conversion steps.

Use case: I am using protobuf encoded messages and encrypt them with
Olm. Now if I want to decrypt an Olm message vodozemac will decrypt the
message convert bytes into string and return it to me. Me, in turn, need
to convert string back into bytes and then decode my protobuf message.
The same thing happens when encoding a message.

What do you think? Current implementation breaks `InboundCreationResult` type for implementation simplicity.

### DIFF
--- a/src/olm/account/mod.rs
+++ b/src/olm/account/mod.rs
@@ -79,7 +79,7 @@ pub struct InboundCreationResult {
     /// The [`Session`] that was created from a pre-key message.
     pub session: Session,
     /// The plaintext of the pre-key message.
-    pub plaintext: String,
+    pub plaintext: Vec<u8>,
 }
 
 /// An Olm account manages all cryptographic keys used on a device.
@@ -235,7 +235,6 @@ impl Account {
 
             // Decrypt the message to check if the Session is actually valid.
             let plaintext = session.decrypt_decoded(&pre_key_message.message)?;
-            let plaintext = String::from_utf8_lossy(&plaintext).to_string();
 
             // We only drop the one-time key now, this is why we can't use a
             // one-time key type that takes `self`. If we didn't do this,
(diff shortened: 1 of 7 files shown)


src/olm/account/mod.rs
```diff
@@ -78,8 +78,15 @@ pub struct IdentityKeys {
 pub struct InboundCreationResult {
     /// The [`Session`] that was created from a pre-key message.
     pub session: Session,
-    /// The plaintext of the pre-key message.
-    pub plaintext: String,
+    /// The plaintext bytes of the pre-key message.
+    pub plaintext_bytes: Vec<u8>,
```
poljar, line 82: Let's keep the `plaintext` name.

src/olm/account/mod.rs
```diff
@@ -78,8 +78,15 @@ pub struct IdentityKeys {
...
+}
+
+impl InboundCreationResult {
+    /// Convert plaintext bytes to string
+    pub fn plaintext(&self) -> String {
```
poljar, line 87: I don't think we need this.

src/olm/session/mod.rs
```diff
@@ -245,12 +245,20 @@ impl Session {
...
     /// [`DecryptionError`]: self::DecryptionError
     pub fn decrypt(&mut self, message: &OlmMessage) -> Result<String, DecryptionError> {
+        self.decrypt_bytes(message).map(|decrypted| String::from_utf8_lossy(&decrypted).to_string())
+    }
+
+    /// Try to decrypt an Olm message, which will either return the plaintext bytes or
+    /// result in a [`DecryptionError`].
+    ///
+    /// [`DecryptionError`]: self::DecryptionError
+    pub fn decrypt_bytes(&mut self, message: &OlmMessage) -> Result<Vec<u8>, DecryptionError> {
```
poljar, line 255: Let's just keep the bytes version of decrypt.

poljar (changes requested):
I was wondering when this will come up, Matrix uses JSON so everything we encrypt/decrypt tends to be a string. So we took a bit of a shortcut here.

The right thing to do is to just accept and returns bytes so I agree with this PR.

Two things before we merge this:
1. Please just leave the byte versions, having both versions, the string version and byte version, of encrypt/decrypt methods makes the  API unnecessarily big.
2. Megolm needs the same treatment.

zaynetro:
Thanks for a review! Reverted the old commit and started from scratch to expose bytes directly.

poljar:
Could you rebase onto latest master to get the CI fixed.

poljar (approved):
Looks good, thanks.

PR #77, opened 2022-07-29
feat(megolm): Add a method to check if two inbound sessions are the same

poljar:


src/megolm/inbound_group_session.rs
```diff
@@ -102,6 +119,58 @@ impl InboundGroupSession {
...
     }
 
+    /// Check if two `InboundGroupSession`s are the same.
+    ///
+    /// An `InboundGroupSession` could be received multiple times with varying
+    /// degreess of trust and first known message indices.
+    ///
+    /// This method allows to check if both `InboundGroupSession`s at the same
+    /// message index provide the same message keys, i.e. if they have been
+    /// indeed created from the same [`GroupSession`].
```
dkasak, line 129:
```suggestion
    /// This method checks if the underlying ratchets of the two
    /// `InboundGroupSession`s are actually the same ratchet, potentially at
    /// a different ratcheting index. That is, if the sessions are *connected*,
    /// then ratcheting one of the ratchets to the index of the other should
    /// yield the same ratchet value, byte-for-byte. This will only be the case
    /// if the `InboundGroupSession`s were created from the same
    /// [`GroupSession`].
```

src/megolm/inbound_group_session.rs
```diff
@@ -102,6 +119,58 @@ impl InboundGroupSession {
...
+    ///
+    /// If the `InboundGroupSession`s are connected, the session with the lower
+    /// message index can safely replace the one with the higher message index.
```
dkasak, line 132:
```suggestion
    /// If the sessions are connected, the session with the lower message index
    /// can safely replace the one with the higher message index.
```

src/megolm/inbound_group_session.rs
```diff
@@ -102,6 +119,58 @@ impl InboundGroupSession {
...
+    pub fn connected(&mut self, other: &mut InboundGroupSession) -> bool {
+        // This method tries to get a `Ratchet` from each session at the same
+        // message index.
```
dkasak, line 135:
```suggestion
        // This method attempts to bring the two `Ratchet` values, one from each
        // session, to the same message index.
```

src/megolm/inbound_group_session.rs
```diff
@@ -102,6 +119,58 @@ impl InboundGroupSession {
...
+            ratchet.ct_eq(&other.initial_ratchet).into()
+        } else if let Some(ratchet) = other.find_ratchet(self.first_known_index()) {
+            self.initial_ratchet.ct_eq(ratchet).into()
+        } else {
+            unreachable!("Either index A >= index B, or vice versa. There is no third option.")
+        }
+    }
+
+    /// Compare the `InboundGroupSession` with the given other
+    /// `InboundGroupSession`
```
dkasak, line 157:
```suggestion
    /// `InboundGroupSession`.
```

dkasak (approved):

PR #81, opened 2022-08-11
Upgrade x25519-dalek to 2.0.0-pre.1

jplatte:
The changes from the currently-used version are [trivial](https://github.com/dalek-cryptography/x25519-dalek/compare/1.2.0...2.0.0-pre.1), but they're required to upgrade to the latest version of zeroize in dependent crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ rand = "0.7.3"
 subtle = "2.4.1"
 sha2 = "0.10.2"
 thiserror = "1.0.30"
-x25519-dalek = { version = "1.2.0", features = ["serde", "reusable_secrets"] }
+x25519-dalek = { version = "2.0.0-pre.1", features = ["serde", "reusable_secrets"] }
 zeroize = "1.3.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"


jplatte:
Oh, looks like `rand` is part of the public API of that crate. I'll look into updating `rand` usage.

poljar:
It is, also that tag/release has been in the `pre` state for quite a while: https://github.com/dalek-cryptography/x25519-dalek/releases/tag/2.0.0-pre.1.

jplatte:
Yeah it has, but I think we're better off with a release that's stable but declared "pre" rathen than a proper release that's constantly causing issues (albeit only at build time).

poljar:
What kind of issues are we talking about?

jplatte:
Not being able to upgrade other dependencies because they require zeroize >=1.4. Also `cargo outdated` doesn't work inside the matrix-sdk workspace, probably because of one such dependency.

jplatte:
*sigh* looks like we'd also need to use https://github.com/dalek-cryptography/ed25519-dalek/pull/160 (through our own fork, I guess). I would still prefer that to the current state of things, but I understand if you don't want to go that route. Let me know if I should continue.

poljar:
> _sigh_ looks like we'd also need to use [dalek-cryptography/ed25519-dalek#160](https://github.com/dalek-cryptography/ed25519-dalek/pull/160) (through our own fork, I guess). I would still prefer that to the current state of things, but I understand if you don't want to go that route. Let me know if I should continue.

I think curve25519-dalek will need to get the same treatment.

I would let that simmer a bit longer, maybe something happens: https://github.com/WebAssembly/wasi-crypto/issues/63.

jplatte:
curve25519-dalek seems to also have a pre-release using the latest version of `rand_core`. So we'd only need to have one fork.

jplatte:
Alright, let's see where this goes: https://github.com/RustCrypto/elliptic-curves/issues/497#issuecomment-1211822232

I'll close this for now.

PR #85, opened 2022-09-05
Add a git-cliff config

poljar:


cliff.toml
```diff
@@ -0,0 +1,65 @@
...
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+    ## [unreleased]
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group | upper_first }}
+    {% for commit in commits %}
+        - {% if commit.breaking %}[**breaking**] {% endif %}{{ commit.message | upper_first }}\
```
dkasak, line 20:
```suggestion
        - {% if commit.breaking %}[**BREAKING**] {% endif %}{{ commit.message | upper_first }}\
```

cliff.toml
```diff
@@ -0,0 +1,65 @@
...
+    { message = "^fix", group = "Bug Fixes"},
+    { message = "^doc", group = "Documentation"},
+    { message = "^perf", group = "Performance"},
+    { message = "^refactor", group = "Refactor"},
+    { message = "^style", group = "Styling"},
+    { message = "^test", group = "Testing"},
+    { message = "^chore\\(release\\): prepare for", skip = true},
+    { message = "^chore", skip = true},
+    { message = "^ci", skip = true},
+    { body = ".*security", group = "Security"},
```
dkasak, line 52: Will this catch `Security`? Presumably not if it's an ordinary regex, so in that case

```suggestion
    { body = ".*[sS]ecurity", group = "Security"},
```

dkasak (approved):
After changes are handled, approved.

cliff.toml
dkasak: Doesn't hurt to have a bit more prominence?

poljar: I don't think I want the changelog to be this shouty.

PR #88, opened 2022-10-07
Add an upgrade method to the InboundGroupSession struct

poljar:


src/megolm/inbound_group_session.rs
```diff
@@ -188,6 +188,54 @@ impl InboundGroupSession {
...
+    /// This method is useful if you receive an `SessionKey` from multiple
+    /// sources with different authentticity properties. For example if you
+    /// receive a `SessionKey` from a fully trusted source with a certain
+    /// ratchet state and an `ExportedSessionKey` from a less trusted source.
+    ///
+    /// You can merge the two sessions into a better one, taking the best
+    /// ratchet state.
+    ///
+    /// Returns `Some` if the sessions could be merged, i.e. they are considered
+    /// to be connected, `None` otherwise.
```
dkasak, line 202:
```suggestion
    /// This method is useful when you receive multiple sessions with
    /// the same session ID but potentially different ratchet indices and
    /// authenticity properties.
    ///
    /// For example, imagine you receive a `SessionKey` S1 with ratchet index
    /// A from a fully-trusted source and an `ExportedSessionKey` S2 with
    /// ratchet state B from a less trusted source. If A > B, then S1 is better
    /// because it's fully trusted, but worse because it's ratcheted further than
    /// S2.
    ///
    /// This method allows you to merge S1 and S2 safely into a fully-trusted S3
    /// with ratchet state B, provided S1 and S2 connect with each other
    /// (meaning they are the same session, just at different ratchet indices).
    ///
    /// Returns `Some(session)` if the sessions could be merged, i.e. they are
    /// considered to be connected and `None` otherwise.
```

dkasak: Attempted to be a bit clearer to someone reading these things for the first time.

dkasak (approved):
Some rewording but otherwise looks good.

dkasak:
Force pushed to fix Conventional Commits stuff.

PR #95, opened 2022-11-30
Support to export into libolm pickles

poljar:


dkasak:
We also need to update the README to remove the notice that libolm pickle support is "read-only".

nico-famedly:
Is there anything blocking this or something I could help to push this forward? I would like to support both libolm and vodozemac for a while in our SDK and if it was possible to switch between both, that would make a few things easier until we migrate away from libolm entirely.

poljar:
> Is there anything blocking this or something I could help to push this forward? I would like to support both libolm and vodozemac for a while in our SDK and if it was possible to switch between both, that would make a few things easier until we migrate away from libolm entirel

I think it's just a lack of time due to a FOSDEM crunch. 

Beware that the transition from libolm to vodozemac is losless, but vodozemac has larger buffers for one-time keys and some other objects.

This makes the transition from vodozemac back to libolm lossy.

src/lib.rs
```diff
@@ -263,9 +263,19 @@ pub enum LibolmPickleError {
...
+    /// libolm pickle. Vodozemac uses an `u64` for its chain index, while libolm
+    /// uses an u32.
+    #[error(
+        "The session contains a chain index that is too big to be put into a libolm pickle: {0}"
+    )]
+    ChainIndexTooBig(u64),
     /// The payload of the pickle could not be decoded.
     #[error(transparent)]
     Decode(#[from] matrix_pickle::DecodeError),
+    /// The object could not been encoded as a pickle.
```
dkasak, line 276:
```suggestion
    /// The object could not be encoded as a pickle.
```

tulir:
Any chance of merging this? I'm thinking of switching mautrix-python to use vodozemac, but the database needs to remain compatible with mautrix-go, which uses libolm or goolm. Longer term goolm could get vodozemac pickling support, but probably won't happen in the near future

poljar:
Let's see if I can resurrect this PR.

poljar:
Alright, the PR is ready for review, that being said:

>  I'm thinking of switching mautrix-python to use vodozemac, but the database needs to remain compatible with mautrix-go, which uses libolm or goolm.

This isn't a particularly great idea, since the conversion to a libolm pickle will always be a lossy operation. One of the reasons for this is because vodozemac can store much more one-time keys compared to libolm.

tulir:
I assume it'll throw away the same keys that libolm would've thrown away at generation time? (as in the behavior is exactly the same either way)

I shouldn't have any need for more than 100 one-time keys nor more than 2 billion messages in an olm session

kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

PR #99, opened 2023-02-27
feat(olm): Add a method to calculate the Session ID from a pre-key message

poljar:
This is useful if we want to log the session ID as soon as we find out we have a pre-key message, in the case of the message failing to decrypt we'll still be able to log the expected session ID.


jplatte (approved):
Apart from the rustdoc warning, looks good 👍🏼

Hywan (approved):

PR #100, opened 2023-04-12
feat!: Return the created and discarded one-time keys when generating new ones

poljar:
The method to generate new one-time keys silently discards old, and creates new one-time keys. It might be useful to the caller to know that some one-time keys were discarded and to see which ones got newly created.

This patch modifies the method `generate_one_time_keys()` so returns a value showing us which, if any, one-time keys got discarded, and which got created.

### DIFF
--- a/src/olm/account/mod.rs
+++ b/src/olm/account/mod.rs
@@ -274,7 +274,6 @@ impl Account {
     }
 
     /// Generates the supplied number of one time keys.
-    ///
     /// Returns the public parts of the one-time keys that were created and
     /// discarded.
     ///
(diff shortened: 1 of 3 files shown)


dkasak (approved):
Looks good.

src/olm/account/mod.rs
```diff
@@ -273,8 +274,20 @@ impl Account {
     }
 
     /// Generates the supplied number of one time keys.
-    pub fn generate_one_time_keys(&mut self, count: usize) {
-        self.one_time_keys.generate(count);
+    ///
```
dkasak, line 277:
```suggestion
```

PR #102, opened 2023-05-11
Bump zeroize

hamza1311:
Bump zeroize dependency version


poljar:
That ain't gonna work until we bump the dalek crates.

hamza1311:
I figured, that's why I closed it...

PR #105, opened 2023-05-16
Return the fallback key if it was removed when generating a new one

poljar:


src/olm/account/mod.rs
```diff
@@ -305,7 +305,10 @@ impl Account {
     ///
     /// The fallback key will be used by other users to establish a `Session` if
     /// all the one-time keys on the server have been used up.
-    pub fn generate_fallback_key(&mut self) {
+    ///
+    /// Returns the public Curve25519 key of a previous fallback key that has
+    /// now been removed from the [`Account`].
```
dkasak, line 310:
```suggestion
    /// Returns the public Curve25519 key of the *previous* fallback key, that
    /// is, the one that will get removed from the [`Account`] when this method
    /// is called.
```

PR #113, opened 2023-09-08
chore: Add a test so we don't regress on accepting padded base64 when decoding

poljar:


src/utilities/mod.rs
```diff
@@ -152,3 +152,17 @@ impl VarInt for u64 {
...
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_b64_decode_any_padding() {
+        let encoded_with_padding = "VGhpc0lzQVRlc3Q=";
+        let encoded_without_padding = "VGhpc0lzQVRlc3Q";
+
+        base64_decode(encoded_with_padding).expect("Should decode if there is padding");
+        base64_decode(encoded_without_padding).expect("Should decode if there is no padding");
```
dkasak, line 166: Maybe also add an assertion that the result is the same.

src/utilities/mod.rs
```diff
@@ -152,3 +152,24 @@ impl VarInt for u64 {
...
+        let encoded_without_padding = "VGhpc0lzQVRlc3Q";
+
+        let first = base64_decode(encoded_with_padding).expect("Should decode if there is padding");
+        let second =
+            base64_decode(encoded_without_padding).expect("Should decode if there is no padding");
+
+        assert_eq!(
+            first,
+            second,
+            "Decoding the same paded base64 string and unpadded base64 string should produce the same result"
```
dkasak, line 172:
```suggestion
            "Decoding the same base64 string with and without padding should produce the same result"
```

dkasak (approved):

PR #114, opened 2023-10-02
Added ExportedSessionKey conversion trait for SessionKey

Michael-Hollister:
Added helper conversion trait for MSC3061 implementation. See following PR for more details: https://github.com/matrix-org/matrix-rust-sdk/pull/2650

Signed-off-by: Michael Hollister <michael@futo.org>

### DIFF
--- a/src/megolm/session_keys.rs
+++ b/src/megolm/session_keys.rs
@@ -314,6 +314,13 @@ impl TryFrom<&str> for SessionKey {
     }
 }
 
+impl From<&ExportedSessionKey> for SessionKey {
+    fn from(value: &ExportedSessionKey) -> Self {
+        let ratchet = Ratchet::from_bytes(value.ratchet.clone(), value.ratchet_index);
+        SessionKey::new(&ratchet, value.signing_key)
+    }
+}
+
 impl Serialize for SessionKey {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where


src/megolm/session_keys.rs
```diff
@@ -314,6 +314,13 @@ impl TryFrom<&str> for SessionKey {
     }
 }
 
+impl From<&ExportedSessionKey> for SessionKey {
```
poljar, line 317: Why is this exactly needed? Why can't you use `InboundGroupSession::import()` with an `ExportedSessionKey`?

poljar (changes requested):
I don't think that this makes much sense, happy to be convinced otherwise though.

src/megolm/session_keys.rs
Michael-Hollister: I made these changes since in a file (`matrix-sdk-crypto/src/gossiping/machine.rs`) where I needed to create a `SessionKey`, the `vodozemac::InboundGroupSession` object was not visible. However after looking again at the available conversion functions in `matrix_sdk_crypto::olm::group_sessions::InboundGroupSession`, I noticed I can instead use the conversion function for `ForwardedMegolmV1AesSha2Content` to create the session, which uses `vodozemac::InboundGroupSession::import()` with the available `ExportedSessionKey`.

I am okay with you closing the PR then since I can make the necessary changes in https://github.com/matrix-org/matrix-rust-sdk/pull/2650. Thanks for checking this out.

poljar:
As discussed, we have a dedicated method to create an `InboundGroupSession` from the `ExportedSessionKey`. Converting from a `SessionKey` to a `ExportedSessionKey` loses some data which we don't want to encourage.

PR #124, opened 2024-02-04
Add interoperability with DR and X3DH based on Signal specs

dkasak:
This implements the delta between Olm and the Signal DR and X3DH specs\[1\]\[2\] as a new mode (dubbed "Interolm"), intended to enable interoperability with libsignal and similar implementations.

\[1\]: https://signal.org/docs/specifications/x3dh/
\[2\]: https://signal.org/docs/specifications/doubleratchet/


dkasak:
This currently breaks libolm session unpickling, because the legacy libolm pickle format doesn't contain some information added by the Interolm mode. This means we have one failing test.

src/olm/messages/message.rs
```diff
@@ -264,6 +389,49 @@ struct ProtoBufMessage {
     ciphertext: Vec<u8>,
 }
 
+#[cfg(feature = "interolm")]
+#[derive(PartialEq, Eq, ProstMessage)]
+struct InterolmProtoBufMessage {
```
clokep, line 394: Seems a bit odd to split `ProtoBufMessage` from its implementation, did you mean for this bit to go below `impl ProtoBufMessage`?

src/olm/messages/message.rs
```diff
@@ -264,6 +389,49 @@ struct ProtoBufMessage {
...
+    previous_counter: u32,
+    #[prost(bytes, tag = "4")]
+    ciphertext: Vec<u8>,
+}
+
+#[cfg(feature = "interolm")]
+impl InterolmProtoBufMessage {
+    const RATCHET_TAG: &'static [u8; 1] = b"\x0A";
+    const INDEX_TAG: &'static [u8; 1] = b"\x10";
+    const PREVIOUS_INDEX_TAG: &'static [u8; 1] = b"\x18";
```
clokep, line 409: Maybe `COUNTER_TAG` and `PREVIOUS_COUNTER_TAG` would match the property names better?

dkasak: We have a mix of the `counter` and `index` nomenclature in various places, including in Megolm (unrelated to this PR). We prefer the `index` terminology, as in the existing Olm implementation, so I've replaced the remaining references to `counter` with it.

The few remaining Megolm locations are best left to a separate PR.

src/olm/messages/message.rs
dkasak: Switched it around, thanks!

poljar:
This PR has now bit-rotted for two years, I think it's time to let it go.

Issue #78, opened 2022-08-01
Mutation testing


poljar:
Mutation testing using [cargo-mutants](https://github.com/sourcefrog/cargo-mutants) reveals a bunch of test cases that aren't checking for all the code behavior we should.

Once we fix all those discovered issues, we should set up `cargo-mutants` to run as a daily CI job:
https://github.com/sourcefrog/cargo-mutants#continuous-integration

- [ ] - Fix issues discovered by mutation testing
- [ ] - Set up cargo-mutant in the CI

Full `cargo-mutant` output follows:
```bash
cargo mutants
Freshen source tree ... ok in 0.051s
Copy source and build products to scratch directory ... 1917 MB in 0.639s
Unmutated baseline ... ok in 13.512s
Auto-set test timeout to 20.0s
Found 393 mutants to test
src/sas.rs:129: replace <impl Debug for EstablishedSas>::fmt -> std::fmt::Result with Ok(Default::default()) ... NOT CAUGHT in 4.381s
src/olm/session/message_key.rs:65: replace MessageKey::key -> &[u8 ; 32] with Default::default() ... NOT CAUGHT in 4.405s
src/olm/session/mod.rs:239: replace Session::next_message_key -> MessageKey with Default::default() ... NOT CAUGHT in 4.225s
src/olm/session/message_key.rs:77: replace MessageKey::index -> u64 with Default::default() ... NOT CAUGHT in 3.954s
src/types/ed25519.rs:276: replace Ed25519Signature::to_base64 -> String with "".into() ... NOT CAUGHT in 3.955s
src/megolm/message.rs:151: replace MegolmMessage::encrypt -> MegolmMessage with Default::default() ... NOT CAUGHT in 3.988s
src/olm/session/receiver_chain.rs:55: replace MessageKeyStore::remove_message_key with () ... NOT CAUGHT in 4.156s
src/olm/session/mod.rs:87: replace ChainStore::is_empty -> bool with true ... NOT CAUGHT in 4.206s
src/olm/account/mod.rs:195: replace Account::remove_one_time_key -> Option<Curve25519SecretKey> with Default::default() ... NOT CAUGHT in 4.182s
src/sas.rs:159: replace SasBytes::decimals -> (u16, u16, u16) with Default::default() ... NOT CAUGHT in 4.275s
src/olm/messages/mod.rs:98: replace OlmMessage::message -> &[u8] with Default::default() ... NOT CAUGHT in 4.140s
[...]

PR #129, opened 2024-03-13
test(olm): Add mutation tests for message_key

Johennes:
Testing the getters is a bit dull but since they're not actually used anywhere else in the library, this seems like the only way to catch their mutants.

I also excluded all `Drop` mutants because it doesn't look like the zeroisation can be tested in Rust. I tried using raw pointers into the struct fields in a test but the memory didn't end up reliably zeroed – I think because `drop` moves the value which can perform a memory copy.

Relates to: #78

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,6 +1,9 @@
+additional_cargo_args = ["--all-features"]
 exclude_re = [
   # Coverage of debug functions is not relevant
   "impl Debug",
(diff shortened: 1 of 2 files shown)


poljar (approved):
Looks good, thanks.

As for the zeroization, we have some test which uses `gdb` here: https://github.com/matrix-org/vodozemac/tree/main/contrib/zeroization-test.

Though it isn't set up to be run as part of CI.

PR #132, opened 2024-03-18
test(olm): Add mutation tests for receiver_chain

Johennes:
This fixes:

```
MISSED   src/olm/session/receiver_chain.rs:58:9: replace MessageKeyStore::remove_message_key with () in 0.8s build + 3.6s test
MISSED   src/olm/session/receiver_chain.rs:122:24: replace > with == in ReceiverChain::find_message_key in 0.8s build + 3.0s test
MISSED   src/olm/session/receiver_chain.rs:135:32: replace - with + in ReceiverChain::find_message_key in 0.8s build + 3.0s test
MISSED   src/olm/session/receiver_chain.rs:135:56: replace > with == in ReceiverChain::find_message_key in 0.9s build + 3.2s test
```

Relates to: #78


src/olm/session/mod.rs
```diff
@@ -599,6 +604,53 @@ mod test {
...
+        );
+
+        // Cannot decrypt first message because it is more than MAX_MESSAGE_KEYS ago
+        assert_matches!(
+            alice_session.decrypt(&messages[0]),
+            Err(DecryptionError::MissingMessageKey(_))
+        );
+
+        // Can decrypt all other messages
+        #[allow(clippy::needless_range_loop)]
```
Johennes, line 629: I believe this isn't a needless range loop because I'm iterating over a subset of the vector only

poljar: What about the suggestion clippy gives us?

```rust
for (i, message) in messages.iter().enumerate().skip(1).take(MAX_MESSAGE_KEYS) {
    assert_eq!(format!("Message {}", i).as_bytes(), alice_session.decrypt(message)?);
}
```

Seems to be a sensible suggestion and you're still iterating over the subset.

src/olm/session/mod.rs
```diff
@@ -599,6 +604,53 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn max_keys_out_of_order_decryption() -> Result<()> {
```
poljar, line 608: I know that we're using a `Result` for a bunch of tests in this repo, but this turns out to produce worse error messages if a test fails, compared to just panicking inside of the test.

Let's convert the `?` calls to `expect()`, or if they are trivial, to `unwrap()` calls.

poljar (changes requested):
Looks good, left some nits.

src/olm/session/mod.rs
Johennes: Sorry, completely missed that. 🤦‍♂️ Now fixed.

poljar (approved):
Nice, thanks for taking care of more of the tests that return a `Result`. Much appreciated.

src/olm/session/mod.rs
Johennes: I have applied this change in two separate commits, one for the new tests I added and one for the existing tests in the same file. Some of the tests have gotten a little more verbose and I'm not sure where exactly to draw the line between using `expect` and `unwrap`. Happy to make further tweaks.

PR #133, opened 2024-03-19
Log ratchet keys in `olm::Session::debug`

### LINKED ISSUE #125: Log the sending chain ratchet key
To build a graph of DH steps we would need to have the ratchet key for the sending chain as well as the receiving chains. That way we'll be able to deduce the sending chain's parent.

More concretely the parent ratchet key of the currently active sending chain would have been the ratchet key kept in the inactive sender chain.

https://github.com/matrix-org/vodozemac/blob/8d0e0f704adeaeacd1cfd66c59ca319e3f1fe016/src/olm/session/double_ratchet.rs#L136-L140

richvdh:
Include the latest public ratchet keys in the debug output.

Fixes #125. Review commit-by-commit.

Question: does any of this need to be behind a feature flag? It's all public data so it feels like it's not a problem to retain or log it, but in theory I suppose there's a reduction in security due to keeping the parent ratchet key around.

### DIFF
--- a/src/olm/session/double_ratchet.rs
+++ b/src/olm/session/double_ratchet.rs
@@ -83,6 +83,7 @@ impl DoubleRatchet {
         let chain_key = ChainKey::new(chain_key);
 
         let ratchet = ActiveDoubleRatchet {
+            parent_ratchet_key: None, // First chain in a session lacks parent ratchet key
             active_ratchet: Ratchet::new(root_key),
             symmetric_key_ratchet: chain_key,
         };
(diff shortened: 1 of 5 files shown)


poljar (changes requested):
Left a small nit, looking good otherwise.

poljar (approved):

poljar:
Do you want to squash that `fixup` commit?

richvdh:
yarp, hold on

poljar:
Ah, now we won't have a changelog entry, I wanted to write a merge commit containing one. :cry:

richvdh:
Oh, sorry!

PR #134, opened 2024-03-19
Track the number of ratchet advances in the Olm Session, and include it in the debug output.

richvdh:
In order to help understand Olm decryption errors, let's keep track of the number of DH ratchet clicks, and log them as part of the debug output.

### DIFF
--- a/src/olm/session/double_ratchet.rs
+++ b/src/olm/session/double_ratchet.rs
@@ -79,6 +79,7 @@ impl DoubleRatchet {
 
         let ratchet = ActiveDoubleRatchet {
             parent_ratchet_key: None, // First chain in a session lacks parent ratchet key
+            ratchet_count: RatchetCount::new(),
             active_ratchet: Ratchet::new(root_key),
             symmetric_key_ratchet: chain_key,
         };
@@ -91,6 +92,7 @@ impl DoubleRatchet {
         Self {
             inner: ActiveDoubleRatchet {
                 parent_ratchet_key: None, // libolm pickle did not record parent ratchet key
+                ratchet_count: RatchetCount::unknown(), // nor the ratchet count
                 active_ratchet: ratchet,
                 symmetric_key_ratchet: chain_key,
             }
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## UNRELEASED
```
poljar, line 5: I'm afraid I need to inform you that vodozemac never abandoned `git-cliff` for changelog entries. So this belongs in a commit message.

poljar (changes requested):
Looks good, I left some nits and I'm going to require tests for this one.

CHANGELOG.md
richvdh: ah, right. Presumably putting it in the subject line of the PR, and then squash-merging the PR, will do the right sort of thing?

src/olm/session/double_ratchet.rs
```diff
@@ -239,9 +278,50 @@ impl Debug for ActiveDoubleRatchet {
...
+            RatchetCount::Known(count) => RatchetCount::Known(count + 1),
+            RatchetCount::Unknown(_) => RatchetCount::Unknown(()),
+        }
+    }
+}
+
+impl Debug for RatchetCount {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RatchetCount::Known(count) => f.write_fmt(format_args!("{}", count)),
```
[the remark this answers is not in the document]
richvdh: I *thought* there must be a better way of doing this!

CHANGELOG.md
poljar: As long as it's a conventional commit it will be picked up, since you historically complained that commit messages don't make great changelog entries you can put the conventional commit into a merge commit.

The merge commit message is anyways useless so it's a good opportunity to repurpose it for a better cause.

In the future I'd like to make it possible so you can have a special tag in merge commits which will mark the start of the changelog entry, similar to how the `Signoff` is treated specially in commit messages.

richvdh:
@poljar: have added some tests. Let me know what you think. Happy to add more if you have suggestions.

poljar (approved):
Looks good, thanks for the tests.

PR #146, opened 2024-04-19
MSC4108 secure channel split encryption keys and monotonic nonces

hughns:
For the attention of @dkasak.

Implement changes from matrix-org/matrix-spec-proposals@aa37af9

### DIFF
--- a/src/secure_channel/mod.rs
+++ b/src/secure_channel/mod.rs
@@ -20,7 +20,7 @@ use thiserror::Error;
 use x25519_dalek::{EphemeralSecret, SharedSecret};
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
-pub use self::messages::{InitialMessage, Message, SecureChannelMessage};
+pub use self::messages::{LoginInitiateMessage, Message, SecureChannelMessage};
 use crate::Curve25519PublicKey;
 
 mod messages;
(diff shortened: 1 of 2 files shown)


src/secure_channel/mod.rs
```diff
@@ -130,27 +130,36 @@ pub struct EstablishedSecureChannel {
...
+    decryption_key: Box<[u8; 32]>,
     #[zeroize(skip)]
     check_code: CheckCode,
+    /// Whether we initiated the secure channel or not. i.e. are we Device G?
+    initiator: bool,
 }
 
+const CHECKCODE_INFO_STRING: &str = "MATRIX_QR_CODE_LOGIN_CHECKCODE";
+const ENCKEY_S_INFO_STRING: &str = "MATRIX_QR_CODE_LOGIN_ENCKEY_S";
+const ENCKEY_G_INFO_STRING: &str = "MATRIX_QR_CODE_LOGIN_ENCKEY_G";
```
poljar, line 145: I like constants to be either completely local, or if they aren't scoped to a `impl` block or even function, at the top of the file. They are now between a struct and its `impl` block which I find to be the last place I would expect them to be.

Though don't worry, I'll move them where I'll want them.

poljar (approved):
Looks like it does what we intended to do. This will go through a more serious review a bit later.

Thanks for tackling this.

PR #142, opened 2024-04-09
test(megolm): Add mutation tests

Johennes:
This fixes the following issues:

```
MISSED   src/megolm/group_session.rs:80:9: replace GroupSession::session_config -> SessionConfig with Default::default() in 0.8s build + 3.1s test
MISSED   src/megolm/inbound_group_session.rs:289:47: replace < with == in InboundGroupSession::find_ratchet in 0.8s build + 3.1s test
MISSED   src/megolm/inbound_group_session.rs:302:9: replace InboundGroupSession::verify_mac -> Result<(), DecryptionError> with Ok(()) in 1.0s build + 3.1s test
MISSED   src/megolm/message.rs:133:9: replace MegolmMessage::add_signature -> Result<(), SignatureError> with Ok(()) in 0.8s build + 3.7s test
MISSED   src/megolm/message.rs:282:42: replace + with - in <impl TryFrom for MegolmMessage>::try_from in 0.8s build + 3.0s test
MISSED   src/megolm/message.rs:50:54: replace + with * in 0.8s build + 3.2s test
MISSED   src/megolm/message.rs:54:9: replace MegolmMessage::ciphertext -> &[u8] with Vec::leak(Vec::new()) in 0.8s build + 3.0s test
MISSED   src/megolm/message.rs:54:9: replace MegolmMessage::ciphertext -> &[u8] with Vec::leak(vec![0]) in 0.8s build + 3.4s test
MISSED   src/megolm/message.rs:54:9: replace MegolmMessage::ciphertext -> &[u8] with Vec::leak(vec![1]) in 0.8s build + 3.0s test
MISSED   src/megolm/message.rs:59:9: replace MegolmMessage::message_index -> u32 with 0 in 0.8s build + 2.9s test
MISSED   src/megolm/message.rs:59:9: replace MegolmMessage::message_index -> u32 with 1 in 0.8s build + 3.0s test
MISSED   src/megolm/message.rs:64:9: replace MegolmMessage::mac -> &[u8] with Vec::leak(Vec::new()) in 0.8s build + 3.1s test
MISSED   src/megolm/message.rs:64:9: replace MegolmMessage::mac -> &[u8] with Vec::leak(vec![0]) in 0.8s build + 3.2s test
MISSED   src/megolm/message.rs:64:9: replace MegolmMessage::mac -> &[u8] with Vec::leak(vec![1]) in 0.9s build + 3.3s test
[...]

PR #147, opened 2024-04-22
test: Enable mutation tests on CI

Johennes:
Depends on: #142
Fixes: #78

This largely follows https://mutants.rs/ci.html and sets up incremental mutation tests on pull requests and full mutation tests on pushes. The full suite takes about half an hour on my laptop.


poljar (approved):
Thanks for taking care of the mutation tests. Very nice work.

Johennes:
Thanks for reviewing all my pull requests! 😅

poljar:
> Thanks for reviewing all my pull requests! 😅

Heh, that wasn't too hard, though took a while because of other responsibilities, sorry about that and thanks again.

PR #150, opened 2024-05-03
chore: Remove the end-to-end-encryption keyword

poljar:
Apparently keywords need to have less than 20 characters.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ description = "A Rust implementation of Olm and Megolm"
 repository = "https://github.com/matrix-org/vodozemac"
 version = "0.6.0"
 edition = "2021"
-keywords = ["matrix", "chat", "messaging", "olm"]
+keywords = ["matrix", "chat", "messaging", "olm", "megolm", "double ratchet", "encryption", "E2EE", "cryptography"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -8,7 +8,7 @@ description = "A Rust implementation of Olm and Megolm"
 repository = "https://github.com/matrix-org/vodozemac"
 version = "0.6.0"
 edition = "2021"
-keywords = ["matrix", "chat", "messaging", "olm", "end-to-end-encryption"]
+keywords = ["matrix", "chat", "messaging", "olm"]
```
dkasak, line 11:
```suggestion
keywords = ["matrix", "chat", "messaging", "olm", "megolm", "double ratchet", "encryption", "E2EE", "cryptography"]
```

dkasak: Should we add a bunch more relevant ones that are < 20 chars while we're at it?

dkasak (approved):
LGTM, but see suggestion.

Cargo.toml
poljar: You can have at most 5 keywords per crate. :face_exhaling:

PR #155, opened 2024-05-17
Clarify status of backup implementation

Johennes:
In a chat with @ara4n yesterday I learnt that the legacy backup has actually been implemented in [matrix-sdk-crypto](https://github.com/matrix-org/matrix-rust-sdk/tree/main/crates/matrix-sdk-crypto/src/backups) and isn't planned to be added here anymore. I suppose [MSC4048](https://github.com/matrix-org/matrix-spec-proposals/pull/4048) will be implemented there, too, then which is why I have removed its reference as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,3 +45,10 @@ Matrix clients, such as [SAS][sas].
 
 [matrix-sdk-crypto]:
 <https://github.com/matrix-org/matrix-rust-sdk/tree/main/crates/matrix-sdk-crypto/src/backups>
+
+## Planned
+
+- Primitives for the asymmetric authenticated [server-side message key backups][authenticated-message-key-backup].
+
+[authenticated-message-key-backup]:
+<https://github.com/matrix-org/matrix-spec-proposals/pull/4048>
--- a/README.md
+++ b/README.md
@@ -37,18 +37,11 @@ Matrix clients, such as [SAS][sas].
 ## Unsupported
 
 - Creating asymmetric [server-side message key
-  backups][legacy-message-key-backup], since they are slated to be enhanced
-  with authentication.
-
-## Planned
-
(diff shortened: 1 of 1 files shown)


dkasak:
We avoided implementing primitives for the existing backup in vodozemac because the design is known to be flawed. Therefore, in my opinion it's not a given that none of the improved design (MSC4048) will end up in vodozemac. The deciding factor for that should be whether something is a "primitive" or foundational cryptographic construct, in which case it belongs in vodozemac, or if it is more of a Matrix layer feature, in which case it should be part of matrix-sdk-crypto.

README.md
```diff
@@ -37,18 +37,11 @@ Matrix clients, such as [SAS][sas].
...
+  backups][legacy-message-key-backup], since these have been implemented in
+  [matrix-sdk-crypto].
 
 [legacy-message-key-backup]:
 <https://spec.matrix.org/v1.2/client-server-api/#server-side-key-backups>
 
-[authenticated-message-key-backup]:
-<https://github.com/matrix-org/matrix-spec-proposals/pull/4048>
+[matrix-sdk-crypto]:
+<https://github.com/matrix-org/matrix-rust-sdk/tree/main/crates/matrix-sdk-crypto/src/backups>
```
dkasak, line 47:
```suggestion
[matrix-sdk-crypto]:
<https://github.com/matrix-org/matrix-rust-sdk/tree/main/crates/matrix-sdk-crypto/src/backups>

## Planned

- Primitives for the asymmetric authenticated [server-side message key backups][authenticated-message-key-backup].

[authenticated-message-key-backup]:
<https://github.com/matrix-org/matrix-spec-proposals/pull/4048>
```

Johennes:
Ah, I see. Thanks for clarifying.

dkasak (approved):

PR #163, opened 2024-07-25
Add a Developer Certificate of Origin

joshsimmons:
The Matrix.org Foundation is rolling out this DCO across all public repositories in its namespace. Of note, this DCO does not require the use of "real" or "legally identifiable" names.


poljar (approved):

poljar:
I wonder if we should enforce this with a GitHub action https://github.com/apps/dco organization-wide.

PR #167, opened 2024-08-02
Apply const keyword to various methods

MTRNord:
This is based on the feedback given by the clippy [`missing_const_for_fn`](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_const_for_fn) lint, which is still in nursery.

In a very non-scientific comparison, the `olm-message-decoding` afl fuzzer runs with ~0.5k to ~1k more executions per second on my PC.

I did not verify if any bindings rely on these to be not `const`. However, the test suite returned that everything is ok.

Signed-Off-By: MTRNord <git@nordgedanken.dev>

### DIFF
--- a/src/cipher/key.rs
+++ b/src/cipher/key.rs
@@ -101,7 +101,7 @@ impl CipherKeys {
         Aes256Key::from_slice(self.aes_key.as_slice())
     }
 
-    pub fn mac_key(&self) -> &HmacSha256Key {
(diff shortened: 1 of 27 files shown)


MTRNord:
(just commeting to ack that I have seen the clippy ci failing. Deal Ing with that when at my pc :))

poljar (approved):
Not sure how terribly useful this is but it shouldn't hurt.

PR #169, opened 2024-08-15
Add a cargo deny configuration

poljar:
This allows us to use a single tool to check for security issues as well as licensing issues and that we don't have any git dependencies.

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -1,0 +1,13 @@
+name: Check for banned licenses
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/deny.yml
```diff
@@ -0,0 +1,13 @@
+name: Check for banned licenses
```
dkasak, line 1:
```suggestion
name: Lint dependencies (for licences, allowed sources, banned dependencies)
```

dkasak (approved):
Looks good modulo a nit.

PR #170, opened 2024-08-15
Update the README

poljar:
Now with a mascot, the image is not ideal considering my ms paint skills but better than nothing.


Hywan (approved):
Zemi ❤️

dkasak (approved):

PR #176, opened 2024-09-05
feat(api)!: Remove base64 expectation for messages in the API

MatMaul:
All the other primitives are dealing directly with bytes I believe so this PR aligns to that.

Signed-off-by: Mathieu Velten <mathieu@velten.xyz>

### DIFF
--- a/src/olm/account/mod.rs
+++ b/src/olm/account/mod.rs
@@ -129,8 +129,8 @@ impl Account {
     }
 
     /// Sign the given message using our Ed25519 fingerprint key.
-    pub fn sign(&self, message: &str) -> Ed25519Signature {
-        self.signing_key.sign(message.as_bytes())
+    pub fn sign(&self, message: impl AsRef<[u8]>) -> Ed25519Signature {
+        self.signing_key.sign(message.as_ref())
     }
 
     /// Get the maximum number of one-time keys the client should keep on the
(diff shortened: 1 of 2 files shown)


src/olm/account/mod.rs
```diff
@@ -129,8 +129,8 @@ impl Account {
     }
 
     /// Sign the given message using our Ed25519 fingerprint key.
-    pub fn sign(&self, message: &str) -> Ed25519Signature {
-        self.signing_key.sign(message.as_bytes())
+    pub fn sign(&self, message: &[u8]) -> Ed25519Signature {
```
poljar, line 132: This could take an `impl AsRef<[u8]>` to avoid a breaking change and to streamline it with `Sesssion::encrypt()`.

poljar (changes requested):
Looks mostly good, but this needs a bit of a different commit message, please take a look at our contributing guide: https://github.com/matrix-org/vodozemac/blob/main/CONTRIBUTING.md#writing-changelog-entries.

Part of it are breaking changes, so make sure to mark them as such using the appropriate git trailer or a `!`in the appropriate place on the first line of the commit message.

poljar (approved):

poljar:
I tweaked the commit message a bit more, it was a bit terse for the changelog, thanks.

PR #177, opened 2024-09-05
fix(doc): Improve appearance of logo against a dark background

andybalaam:
The border of the Zemi image faded out to white, which left a halo when viewed against a dark background. This change makes it fade out to transparent.

Before:

image

After:

image


poljar (approved):
Nice.

PR #185, opened 2024-09-19
Fill out the remaining missing docs and configure denying of missing docs from now on

poljar:


src/olm/account/mod.rs
```diff
@@ -113,17 +115,17 @@ impl Account {
...
     }
 
-    /// Get a reference to the account's public Ed25519 key
+    /// Get a copy of the account's public Ed25519 key
     pub const fn ed25519_key(&self) -> Ed25519PublicKey {
         self.signing_key.public_key()
     }
 
-    /// Get a reference to the account's public Curve25519 key
+    /// Get a copy to the account's public Curve25519 key
```
andybalaam, line 128: copy to -> copy of

andybalaam (approved):
Looks good, one tiny change.

src/types/mod.rs
```diff
@@ -35,28 +39,51 @@ impl From<KeyId> for String {
...
     Signature(#[from] SignatureError),
-    /// At least one of the keys did not have contributory behaviour and the
-    /// resulting shared secret would have been insecure.
+
+    /// One or more keys lacked contributory behavior in the Diffie-Hellman
+    /// operation, resulting in an insecure shared secret.
+    ///
+    /// For more details on contributory behavior please refer to the
+    /// [`x25519_dalek::SharedSecret::was_contributory()`] method.
     #[error("At least one of the keys did not have contributory behaviour")]
```
dkasak, line 87: Should probably change the error to use the same phrasing?

dkasak (approved):

src/olm/account/mod.rs
poljar: 50b8fd3cbeb8bda10bf3cf0d19904fbe7fbc0b5d

src/types/mod.rs
poljar: d9e9a70c6ddf3e8bd6b2f6021533806e4ce47aa3

src/cipher/mod.rs
```diff
@@ -83,28 +89,53 @@ pub enum DecryptionError {
...
+    ///
+    /// This key derivation format is typically used for generating individual
+    /// message keys in the Megolm ratchet.
     pub fn new_megolm(&key: &[u8; 128]) -> Self {
         let keys = CipherKeys::new_megolm(&key);
 
         Self { keys }
     }
 
+    /// Creates a new [`Cipher`] from the given 32-byte raw key.
```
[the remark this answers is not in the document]
poljar: Well it should™ be a 32-byte key, but libolm accepted any length so here we are.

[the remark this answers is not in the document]
poljar: 0d528765708ebfee1250a5e175893c619b1ae866

PR #194, opened 2024-10-08
Fix the zeroization test

poljar:
The test used some internal API to test if our zeroization works. We have now switched to the publicly available Ed25519 secret key type.

Some additional changes were needed to adjust for the different sizes of the secret key material.


poljar:
The test fails, but CI did not notice this because the make file eats the error code.

Not sure why the test fails since it works locally.

kathleenmotley49-maker (approved):

PR #195, opened 2024-10-09
chore: Allow the usage of the breaking-change trailer for changelog entries

poljar:


CONTRIBUTING.md
```diff
@@ -40,7 +40,8 @@ olm, cipher).
 In addition to the Conventional Commit format, you can use the `Changelog` git
 trailer to specify the changelog message explicitly. When that trailer is
 present, its value will be used as the changelog entry instead of the commit's
-leading line.
+leading line. The `Breaking-Change` git trailer can be used in a similar manner
+if the changelog entry should be marked as a breaking change.
```
dkasak, line 44:
```suggestion
leading line. For breaking changes, use the `Breaking-Change` trailer instead of the `Changelog` trailer to specify the changelog entry.
```

dkasak: I'd suggest this to make it clear that `Breaking-Change` simply replaces `Changelog` in case of a breaking change.

dkasak (approved):

Issue #196, opened 2024-10-30
Dehydrated Device: Serialize and deserialize new format


BillCarsonFr:
Serialize and deserialize the format defined in https://github.com/element-hq/element-meta/issues/2602 When deserializing and creating a new Account, some fields may need to be populated with fake values if they are missing from the dehydrated device format.

PR #199, opened 2024-11-27
Add support for new dehydrated device format

uhoreg:
as documented in the current version of [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814)

Fixes https://github.com/matrix-org/vodozemac/issues/196


poljar:
Gonna close and reopen this to see if the codecov thing kicks in.

uhoreg:
@poljar can I get a review on this PR?  Thanks

src/olm/account/mod.rs
```diff
@@ -682,16 +721,167 @@ mod libolm {
...
+            })
+        }
+    }
+
+    pub fn expand_pickle_key(key: &[u8; 32], identity_key: &str) -> Box<[u8; 32]> {
+        let kdf: Hkdf<Sha256> = Hkdf::new(Some(identity_key.as_bytes()), key);
+        let mut key = [0u8; 32];
+
+        kdf.expand(b"dehydrated-device-pickle-key", &mut key)
+            .expect("We should be able to expand the 32 byte pickle key");
```
poljar, line 868: Ah, we do avoid the IV reuse. Why is the identity key passed as a string here? We should probably concatenate the identity key to the info string instead of putting it into the salt, a mistake I made in the SDK a well.

Though a random IV/nonce would have been even better, this would avoid the IV reuse even if we re-upload the same pickle with a fresh set of one-time keys or for whatever other reason we might want to re-encrypt the dehydrated device pickle.

poljar (changes requested):
Hi, sorry for the late reply, we miscommunicated on who would do the first review.

A good start but needs a bit more polish.

src/olm/account/mod.rs
```diff
@@ -439,6 +439,45 @@ impl Account {
...
+        curve25519key: &str,
+        key: &[u8; 32],
+    ) -> Result<Self, crate::LibolmPickleError> {
+        use self::dehydrated_device::{expand_pickle_key, Pickle, PICKLE_VERSION};
+        use crate::utilities::unpickle_libolm;
+
+        let pickle_key = expand_pickle_key(key, curve25519key);
+
+        #[cfg(feature = "libolm-compat")]
+        return unpickle_libolm::<Pickle, _>(pickle, pickle_key.as_ref(), PICKLE_VERSION).or_else(
```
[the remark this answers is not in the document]
uhoreg: I had tried that, but the compiler doesn't like it because it thinks it isn't a final statement (even though this `#[cfg]` and the next one are mutually exclusive).

src/olm/account/mod.rs
```diff
@@ -439,6 +439,45 @@ impl Account {
 
         pickle.try_into()
     }
+
+    /// Create a dehydrated device pickle.
```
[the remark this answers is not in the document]
uhoreg: Yup, I had considered fixing the IV problem, but also wasn't sure if I wanted to introduce a new encryption scheme into the code.  I'll think about it some more.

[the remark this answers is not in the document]
poljar: We now have a dep on `Chacha20` in vodozemac due to the QR Login work, and it's likely going to be needed for the PQ support.

So using a new scheme isn't as problematic anymore.

src/olm/account/mod.rs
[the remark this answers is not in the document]
uhoreg: Since I'm going to change the encryption, it probably makes sense to deal with compatibility in the SDK layer rather than here, so the `#[cfg]` will go away.

src/olm/account/mod.rs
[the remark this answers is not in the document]
uhoreg: Added more to the comment, and switch encryption to using Chacha20poly1305

src/olm/account/mod.rs
```diff
@@ -439,6 +439,45 @@ impl Account {
...
+
+        let pickle_key = expand_pickle_key(key, &self.curve25519_key().to_base64());
+
+        pickle_libolm::<Pickle>(self.into(), pickle_key.as_ref())
+    }
+
+    /// Create an [`Account`] object by rehydrating a device.
+    pub fn from_dehydrated_device(
+        pickle: &str,
+        curve25519key: &str,
```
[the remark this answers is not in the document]
uhoreg: Changed parameters, so this parameter is gone now.

src/olm/account/mod.rs
```diff
@@ -682,16 +721,167 @@ mod libolm {
...
+
+                    ret
+                }
+            };
+
+            Ok(ret)
+        }
+    }
+
+    #[derive(Encode, Decode, Zeroize, ZeroizeOnDrop)]
```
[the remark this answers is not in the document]
uhoreg: done

src/olm/account/mod.rs
```diff
@@ -682,16 +721,167 @@ mod libolm {
...
+                .map(|(_key_id, secret_key)| OneTimeKey { private_key: secret_key.to_bytes() })
+                .collect();
+
+            let fallback_key =
+                account.fallback_keys.fallback_key.as_ref().and_then(|f| f.try_into().ok());
+
+            Self {
+                version: PICKLE_VERSION,
+                private_curve25519_key: account.diffie_hellman_key.secret_key().to_bytes(),
+                private_ed25519_key: account.signing_key.expanded_secret_key(),
```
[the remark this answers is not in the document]
uhoreg: fixed

src/lib.rs
```diff
@@ -262,6 +261,53 @@ pub enum LibolmPickleError {
...
+        match err {
+            LibolmPickleError::PublicKey(e) => Self::PublicKey(e),
+            LibolmPickleError::Decode(e) => Self::Decode(e),
+            LibolmPickleError::Encode(e) => Self::Encode(e),
+            _ => panic!("Unexpected libolm pickle error {}", err),
+        }
+    }
+}
+
+impl From<chacha20poly1305::aead::Error> for DehydratedDeviceError {
```
uhoreg, line 305: I tried using `Decryption(#[from] chacha20poly1305::aead::Error)`, but it was complaining about some missing traits.  So I had to do this instead.  I'm not sure if there's a better way.

src/olm/account/mod.rs
```diff
@@ -439,6 +444,79 @@ impl Account {
...
+    ///
+    /// Returns the ciphertext and nonce.  `key` is a 256-bit (32-byte) key for
+    /// encrypting the device.
+    ///
+    /// The format used here is defined in
+    /// [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814).
+    pub fn to_dehydrated_device(
+        &self,
+        key: &[u8; 32],
+    ) -> Result<(String, String), crate::DehydratedDeviceError> {
```
uhoreg, line 464: Should I create a new struct instead of returning `(String, String)`?

src/olm/account/mod.rs
```diff
@@ -682,16 +721,167 @@ mod libolm {
...
+    }
+
+    #[derive(Encode, Decode, Zeroize, ZeroizeOnDrop)]
+    pub(super) struct Pickle {
+        version: u32,
+        private_curve25519_key: Box<[u8; 32]>,
+        private_ed25519_key: Box<[u8; 64]>,
+        one_time_keys: Vec<OneTimeKey>,
+        opt_fallback_key: OptFallbackKey,
+    }
```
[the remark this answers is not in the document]
uhoreg: Forgot to respond to this earlier.

I've written the basic code:

```rust
use afl::fuzz;
use arbitrary::Arbitrary;
use vodozemac::olm::Account;

#[derive(Arbitrary)]
struct Data {
    ciphertext: String,
    nonce: String,
    key: [u8; 32],
}

fn main() {
    fuzz!(|data: Data| {
        let _ = Account::from_dehydrated_device(&data.ciphertext, &data.nonce, &data.key);
    });
}
```

It looks like the afl documentation is suggesting to provide some example input? Or can I just let it provide random input?

To use it, do I just run `cargo afl fuzz`?  When I do that, I get a permission error -- which may be because my home directory is mounted `noexec`, so I need to figure out how to get it to run.  But I was wondering if there was something else I need to add to the command line.

src/olm/account/mod.rs
poljar: Yeah that sounds good, and it might implement `Serialize`

uhoreg:
The clippy failure seems to be in code that I did not change.  I'm guessing that I should create a new PR to fix that, rather than adding the fixes to this PR.

poljar:
> The clippy failure seems to be in code that I did not change. I'm guessing that I should create a new PR to fix that, rather than adding the fixes to this PR.

Should be fixed in this PR: https://github.com/matrix-org/vodozemac/pull/200.

src/olm/account/mod.rs
```diff
@@ -439,6 +453,85 @@ impl Account {
...
+        let pickle: Pickle = self.into();
+        let mut encoded = pickle.encode_to_vec()?;
+        let cipher = ChaCha20Poly1305::new(key.into());
+        let rng = thread_rng();
+        let nonce = ChaCha20Poly1305::generate_nonce(rng);
+        let ciphertext = cipher.encrypt(&nonce, encoded.as_slice());
+        encoded.zeroize();
+        let ciphertext = ciphertext?;
+
+        Ok(DehydratedDeviceResult {
```
poljar, line 491: Hmm, I know I said a struct would be nice, but perhaps we should encode it using `MatrixPickle` and then base64 encode it, though that requires again a version field.

We would have an inner pickle and an outer pickle, the inner pickle is the unencrypted account and the outer pickle is the ciphertext + nonce.

It's quite similar to the Olm/Megolm message packages, though there we use protobuf, so we could even pick that for consistency. What do you think about that?

uhoreg: Yeah, I had considered that.  I prefer to have the data format be a standard format as much as possible, rather than increasing the number of custom binary formats that we have.  So I figured it would be better to return the nonce and ciphertext separately, and allow the SDK layer to put them in the JSON.

src/olm/account/mod.rs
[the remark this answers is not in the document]
uhoreg: > > When I do that, I get a permission error -- which may be because my home directory is mounted noexec, so I need to figure out how to get it to run.
> 
> Yeah that might be problematic, does this mean that you can't run examples/tests either?

Tests work for me.  I haven't tried examples.  I worked around the issue by running it in Docker.

> I don't think you are targeting the correct thing here, you would target the unencrypted pickle. The pickle gets decoded from a byte array so you would then have no trouble providing it raw bytes as an input.

Right.  I wasn't sure if I should be copying the pickle examples, or something like `olm-decryption`.  But I've change it now.  I also created two sample input files, which are just (unencrypted) pickles from new accounts, one with 10 OTKs and a fallback key, and one with 20 OTKs and no fallback key.  Because afl didn't want to run without any input files.

src/lib.rs
```diff
@@ -262,6 +261,49 @@ pub enum LibolmPickleError {
...
+
+impl From<LibolmPickleError> for DehydratedDeviceError {
+    fn from(err: LibolmPickleError) -> Self {
+        match err {
+            LibolmPickleError::PublicKey(e) => Self::PublicKey(e),
+            LibolmPickleError::Decode(e) => Self::Decode(e),
+            LibolmPickleError::Encode(e) => Self::Encode(e),
+            // Any other errors should not happen when we're dealing with
+            // dehydrated devices.
+            _ => panic!("Unexpected libolm pickle error {}", err),
```
poljar, line 302: No panics please, just include the full error type in one of the variants of `DehydratedDeviceError`.

src/olm/account/mod.rs
```diff
@@ -439,6 +453,102 @@ impl Account {
...
+            return Err(crate::DehydratedDeviceError::InvalidNonce);
+        }
+        let mut plaintext = cipher.decrypt(nonce.as_slice().into(), ciphertext.as_slice())?;
+        let version =
+            get_pickle_version(&plaintext).ok_or(crate::DehydratedDeviceError::MissingVersion)?;
+        if version != PICKLE_VERSION {
+            return Err(crate::DehydratedDeviceError::Version(PICKLE_VERSION, version));
+        }
+
+        let mut cursor = Cursor::new(&plaintext);
```
poljar, line 528: Can't the `from_decrypted_dehydrated_device()` function be used here? Otherwise we're fuzzing a function that we're not actually using.

src/olm/account/mod.rs
```diff
@@ -682,16 +792,163 @@ mod libolm {
...
+        Account,
+    };
+    use crate::{
+        types::{Curve25519Keypair, Curve25519SecretKey},
+        DehydratedDeviceError, Ed25519Keypair, KeyId,
+    };
+
+    #[derive(Encode, Decode, Zeroize, ZeroizeOnDrop)]
+    pub(crate) struct OneTimeKey {
+        pub(crate) private_key: Box<[u8; 32]>,
```
poljar, line 811: I know this is a bit useless since you are boxing the value, but we should use the `#[secret]` annotation anyways: https://github.com/matrix-org/matrix-pickle/?tab=readme-ov-file#encoding-and-decoding-secrets.

src/olm/account/mod.rs
```diff
@@ -682,16 +792,163 @@ mod libolm {
...
+        type Error = ();
+
+        fn try_from(key: &FallbackKey) -> Result<Self, ()> {
+            Ok(OneTimeKey { private_key: key.secret_key().to_bytes() })
+        }
+    }
+
+    #[derive(Zeroize, ZeroizeOnDrop)]
+    pub(crate) struct OptFallbackKey {
+        pub(crate) fallback_key: Option<OneTimeKey>,
```
poljar, line 834: Not something for now, but perhaps it'll make sense to implement `Decode` and `Encode` support for `Option<T>` in `matrix-pickle`.

src/olm/account/mod.rs
```diff
@@ -1197,4 +1454,149 @@ mod test {
...
+        let carol_olm_message = carol_session.encrypt(message);
+
+        let mut alice_rehydrated = Account::from_dehydrated_device(
+            &alice_dehydrated_result.ciphertext,
+            &alice_dehydrated_result.nonce,
+            &PICKLE_KEY,
+        )
+        .expect("Should be able to rehydrate device");
+
+        if let OlmMessage::PreKey(m) = bob_olm_message {
```
poljar, line 1503: We use `assert_matches!()` or `assert_let` for this construction nowadays:

```rust
let prekey_message = assert_matches!(olm_message, OlmMessage::PreKey(m) => m);
```

poljar (changes requested):
Found a couple of minor things, but I think after that it's good to go.

I'm fuzzing the new target so we have a large corpus that we can run as part of the unit tests (We do that for every fuzz target).

src/olm/account/mod.rs
uhoreg: Yup, I was considering that, but figured it would be easier to do it here for now.  Maybe the next time we need to support an `Option<T>`.

uhoreg:
I think that the clippy errors are (again) unrelated to this PR

poljar:
> I think that the clippy errors are (again) unrelated to this PR

Yeah, that lint might not have been the best idea: https://github.com/matrix-org/vodozemac/pull/203.

poljar:
Rebased to get the CI unblocked.

src/lib.rs
poljar: This has been removed.

src/olm/account/mod.rs
poljar: This has been implemented.

src/olm/account/mod.rs
poljar: I added those myself.

src/olm/account/mod.rs
```diff
@@ -682,16 +790,171 @@ mod libolm {
...
+            let fallback_key =
+                account.fallback_keys.fallback_key.as_ref().and_then(|f| f.try_into().ok());
+
+            Self {
+                version: PICKLE_VERSION,
+                private_curve25519_key: account.diffie_hellman_key.secret_key().to_bytes(),
+                private_ed25519_key: account
+                    .signing_key
+                    .unexpanded_secret_key()
+                    .expect("Cannot dehydrate an account created from a libolm pickle"),
```
poljar, line 904: Actually, @uhoreg, won't this pose a problem? In any case I don't think we should use `expect()` here.

poljar: So the expectation is that we ourselves will always create a new Account when we want to create a dehydrated device, but people may misuse this.

I think we should convert this into a `TryFrom` and let the person be aware of this limitation. In the Rust SDK itself we can guarantee that we don't misuse the API and call `expect()` there.

PR #204, opened 2025-01-17
Don't panic if we try to dehydrate an Account libolm has created

poljar:

PR #205, opened 2025-01-17
Disallow panics, unreachables, and expects using clippy lints

poljar:
Now that we're relatively stable, we can disallow the usage of these quite dangerous invocations.

We don't want to let any of those slip through reviews and it allows people that write PRs know ahead of time that they shouldn't be used.

Depends on #204 since that removes a panic that shouldn't exist.


poljar:
Rebased now that #204 has been merged.

dkasak (approved):
LGTM

PR #209, opened 2025-05-06
chore: Allow the usage of a git source for olm-sys

poljar:
This was necessary because olm-sys started to fail to build because of the minimally required CMake version needed to be bumped.

### DIFF
--- a/src/olm/account/mod.rs
+++ b/src/olm/account/mod.rs
@@ -896,8 +896,8 @@ mod dehydrated_device {
             let one_time_keys: Vec<_> = account
                 .one_time_keys
                 .secret_keys()
-                .iter()
-                .map(|(_key_id, secret_key)| OneTimeKey { private_key: secret_key.to_bytes() })
+                .values()
+                .map(|secret_key| OneTimeKey { private_key: secret_key.to_bytes() })
                 .collect();
 
             let fallback_key =
--- a/src/olm/session/mod.rs
+++ b/src/olm/session/mod.rs
@@ -662,7 +662,7 @@ mod test {
 
         let mut messages: Vec<messages::OlmMessage> = Vec::new();
         for i in 0..(MAX_MESSAGE_KEYS + 2) {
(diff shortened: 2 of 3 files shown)


src/olm/account/mod.rs
```diff
@@ -896,8 +896,8 @@ mod dehydrated_device {
             let one_time_keys: Vec<_> = account
                 .one_time_keys
                 .secret_keys()
-                .iter()
-                .map(|(_key_id, secret_key)| OneTimeKey { private_key: secret_key.to_bytes() })
+                .values()
```
andybalaam, line 899: Is this related to the olm-sys change?

src/olm/session/mod.rs
```diff
@@ -662,7 +662,7 @@ mod test {
 
         let mut messages: Vec<messages::OlmMessage> = Vec::new();
         for i in 0..(MAX_MESSAGE_KEYS + 2) {
-            messages.push(bob_session.encrypt(format!("Message {}", i).as_str()).into());
+            messages.push(bob_session.encrypt(format!("Message {i}").as_str()).into());
```
andybalaam, line 665: This one definitely looks unrelated.

For absolute clarity, I'd suggest making this a separate commit.

andybalaam (commented):
Looks good, but I'd prefer to separate out unrelated changes.

src/olm/session/mod.rs
poljar: But it is a separate commit?

src/olm/account/mod.rs
poljar: It's not, the commit tells you that it's a new clippy warning that is being fixed.

andybalaam: Sorry, I assumed the clippy warnings were related to the changes made in the first commit.

andybalaam (approved):
Separate changes are already in separate commits, I misread the diff. Looks good!

PR #211, opened 2025-05-29
Add some more tests to catch some new mutations cargo mutants produces

### LINKED ISSUE #210: Fix some tests that produce failures in the mutation testing run
Seems that some new tests fail to produce an error if the code under the test is mutated.

Nothing major, but it produces a CI failure. Example test run is found here: https://github.com/matrix-org/vodozemac/actions/runs/14903389002/job/41860202664#step:6:350

poljar:
This is mostly for debug implementations but there were some interesting cases as well.

Review commit by commit is recommended.

This closes #210.


andybalaam (approved):
Wow, `assert_ne` is very similar to `assert_eq`! If there were a way of making it easier to see that would be great but I can't immediately think of a very good way.

dkasak (approved):

PR #213, opened 2025-06-20
Switch to llvm-cov for coverage reports

poljar:


justfile
Hywan, line 1: Hurray for `justfile` \o/

Hywan (approved):

richvdh:
Could you give some background to this, for the archive? I gather we've been having problems with tarpaulin?

justfile
```diff
@@ -0,0 +1,13 @@
...
+
+# List the available recipes.
+default:
+    just --list
+
+# Create a coverage report using llvm-cov.
+coverage:
+    cargo llvm-cov nextest
+
+# Create a coverage report for codecov.
```
richvdh, line 11: > # Create a coverage report using llvm-cov.
> # Create a coverage report for codecov.

one of these says which tool it uses to *make* the report (`llvm-cov`) while the other says which tool the report is *for* (`codecov`). Don't they both use `llvm-cov`? If the report generated by `coverage` is not for `codecov`, what is it for?

richvdh: Or drop the "using llvm-cov" from both if you think it's unimportant

poljar: I think it's fine to mention it, as the tool has many more options. The justfile is here to more easily use the default flows and aid in the discoverability of the coverage report support we have.

.github/workflows/coverage.yml
```diff
@@ -22,20 +22,31 @@ jobs:
...
-    - name: Install tarpaulin
-      run: cargo install cargo-tarpaulin
+    - name: Install cargo-llvm-cov
+      uses: taiki-e/install-action@cargo-llvm-cov
+
+    - name: Install nextest
+      uses: taiki-e/install-action@nextest
+
+    - name: Install just
+      uses: taiki-e/install-action@just
```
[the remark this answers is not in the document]
poljar: Did so in https://github.com/matrix-org/vodozemac/pull/213/commits/2961469ff11996353c838f09f9e3cf582ed103c5.

PR #217, opened 2025-07-03
Address a bunch of mutation test failures

poljar:
I couldn't figure out how to tests some of those, and some added tests aren't quite pretty, but most of them should be fine.

A review commit by commit is probably the best.

This closes: https://github.com/matrix-org/vodozemac/issues/210


src/types/curve25519.rs
```diff
@@ -244,6 +244,7 @@ impl From<Curve25519Keypair> for Curve25519KeypairPickle {
 
 #[cfg(test)]
 mod tests {
+    use assert_matches2::assert_matches;
```
dkasak, line 247: Why do we need `assert_matches2` here? Or is the `2` a typo?

src/megolm/message.rs
```diff
@@ -354,6 +354,8 @@ mod test {
     #[cfg(feature = "low-level-api")]
     use std::vec;
 
+    use assert_matches2::assert_matches;
```
dkasak, line 357: Similarly, why do we need `assert_matches2` here?

src/megolm/inbound_group_session.rs
```diff
@@ -706,4 +706,73 @@ mod test {
...
+        assert_eq!(ratchet.index(), 2);
+        assert_ne!(address, initial_address);
+        assert_eq!(
+            address, latest_ratchet,
+            "We should have advanced and returned the latest ratchet"
+        );
+
+        let ratchet = session
+            .find_ratchet(2)
+            .expect("We should be able to find the ratchet at message index 2");
```
dkasak, line 756:
```suggestion
            .expect("We should be able to find the ratchet at message index 2 yet again");
```

src/megolm/inbound_group_session.rs
```diff
@@ -706,4 +706,73 @@ mod test {
...
+        let address = ratchet.ratchet_bytes_pointer();
+
+        assert_eq!(ratchet.index(), 2);
+        assert_ne!(address, initial_address);
+        assert_eq!(
+            address, latest_ratchet,
+            "We should have advanced and returned the latest ratchet"
```
dkasak, line 763:
```suggestion
            "We asked for the same ratchet index so we should again get the same latest ratchet"
```

dkasak (changes requested):

src/megolm/ratchet.rs
```diff
@@ -285,6 +290,34 @@ mod tests {
...
+                3, 7, 119, 230, 128, 217, 229, 92, 83, 216, 5, 131, 22, 188, 159, 163, 79, 138,
+                227, 80, 19, 63, 163, 36, 116, 128, 80, 109, 235, 104, 121, 44, 74, 58, 219, 178,
+                203, 250, 94, 173, 74, 141, 82, 227, 6, 151, 162, 53, 4, 239, 5, 245, 68, 206, 102,
+                23, 30, 198, 227, 138, 58, 111, 173, 151, 116, 218, 128, 12, 49, 14, 93, 134, 249,
+                118, 99, 16, 135, 52, 150, 192, 182, 48, 229, 161, 247, 35, 208, 27, 40, 157, 250,
+                63, 114, 74, 207, 133, 188, 44, 115, 220, 157, 154, 101, 220, 79, 250, 24, 205, 26,
+                90
+            ]
+        );
+    }
```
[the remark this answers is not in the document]
poljar: > @poljar says we actually prevent calling advance_to with older values in the higher-level API (Rust SDK crypto crate?) so maybe we need to rip out the code that tries handling this case instead?

It's right here https://github.com/matrix-org/vodozemac/blob/7a91e0003fbc4bfe48f14ffe7dd77c2d460da96b/src/megolm/inbound_group_session.rs#L318-L333

We only call `advance_to()` if the current index is smaller than the target index.

> I find it concerning that we don't actually understand the purpose of this "fall back to 256 steps" code and, consequently, that this test is enshrining that behavior. I don't see what's stopping the caller from corrupting the ratchet state if they accidentally call advance_to with an old value. In fact, that seems to be exactly what this test is testing for, that the ratchet state is indeed corrupted in that scenario.

Well sure, but this has been sadly ported from libolm verbatim and there are plans to migrate Megolm to a simpler ratchet. We can also ignore this mutation if you prefer that. But the tests will at least warn us if the behavior here changes, which sounds useful on its own.

PR #220, opened 2025-07-18
Set up cargo-hack

poljar:
This allows us to check that our features work independently of each other.

This also includes a fix for one of our features that indeed did not work without another feature being enabled.


andybalaam (approved):
Looks good and useful to me.

PR #224, opened 2025-07-24
DO NOT MERGE: Introduce a performance regression

poljar:


poljar:
> ## [CodSpeed Performance Report](https://codspeed.io/matrix-org/vodozemac/branches/poljar%2Ftest-benchmarks)
> ### Merging #224 will **not alter performance**
> 
> Comparing `poljar/test-benchmarks` ([df0dda4](https://github.com/matrix-org/vodozemac/commit/df0dda489eab00e43c152de59d62c7fbe02a96d7)) with `main` ([12588c5](https://github.com/matrix-org/vodozemac/commit/12588c5a66ab9e81305b1611aa6741ab42c7dd7a))
> ### Summary
> 
> `✅ 3` untouched benchmarks

A bit worrying that the bot claims this, looking at the codspeed dashboard it's ignoring the sleep because it's a system call:

> This benchmark contains 1 system calls, totalling 1.1 ms of execution time. Since they cannot be consistently instrumented, those calls are not included in the measure. Please switch to [the Walltime instrument to accurately measure system calls](https://codspeed.io/docs/instruments/walltime). Learn more [about measurement and system calls](https://codspeed.io/docs/instruments/cpu/overview#system-calls).

There's an option to view the how the performance changed including system calls and there it does warn us that we regressed. It also shows which function calls have been added since the last benchmark run.

image

All the blue pills in the stack are newly added calls.

Pretty neat I would say.

PR #246, opened 2025-09-13
Add missing to_libolm_pickle implementations

tulir:
I don't know Rust and haven't tested this yet


tulir:
Oh #95 does this already, it just doesn't show up when searching for to_libolm_pickle 🙃

PR #281, opened 2026-01-12
Update some deps

poljar:
Mainly to get a newer cargo deny and to bump our criterion version which depends on the now unmaintained bincode.


bnjbvr (approved):
Simple enough, lgtm, thanks!

PR #287, opened 2026-01-12
fix: default SessionConfig to version 1

poljar:
The default() implementations of olm::SessionConfig and megolm::SessionConfig now create a version 1 SessionConfig instead of version 2.

Version 2 isn't specified anywhere yet, it shouldn't be used outside of experimental setups.

This closes https://github.com/matrix-org/vodozemac/issues/280.


richvdh (approved):
seems sensible to me.

PR #288, opened 2026-01-13
test: Add more tests for our MAC verification logic

poljar:
This became necessary as we merged #287, now we're missing a test for the MAC verification of the untruncated MAC.

Link to the mutation test that is now failing: https://github.com/matrix-org/vodozemac/actions/runs/20922478894/job/60111402253.


andybalaam (approved):
Looks good, thanks!

PR #322, opened 2026-04-08
Bump some deps and update the contributing guide

poljar:
We never updated the contributing guide when we switched to manually written changelog entries.

So let's do that now. There's some miscellaneous cleanup as well, i.e. the tarpaulin config isn't relevant anymore since we're using `llvm-cov`.


andybalaam (approved):
Looks good.

I notice we ignore Cargo.lock - might we want to revisit that decision following https://blog.rust-lang.org/2023/08/29/committing-lockfiles/ or has this already been thought about?

PR #341, opened 2026-05-08
Hazmat: expose Session constructor from non-Olm-derived root key material

### LINKED ISSUE #334: Hazmat: expose Session constructor from non-Olm-derived root key material
## Motivation

Vodozemac's Double Ratchet implementation is production-quality and Cure53-audited. Currently `Session` construction requires going through `Account` and Olm's 3DH handshake, and the `from_ratchet_and_chain_key` side-door is gated on `libolm-compat` — implying the maintainers already accept this reuse pattern for compatibility purposes.

I'd like to propose extending that reuse pattern to new protocol designs: a hazmat-feature-gated `Session::from_root_key_material` + a companion `Session::active_sending_state` accessor (the latter strictly for equivalence-testing against `Account`-derived sessions), enabling downstream projects to reuse vodozemac's DR over non-Olm handshakes (Noise variants, alternative PQ KEMs, experimental transports).

## Proposal

Behind a new opt-in feature `hazmat-expose-root-key`:

```rust
#[cfg(feature = "hazmat-expose-root-key")]
impl Session {
    pub fn from_root_key_material(
        config: SessionConfig,
        session_keys: SessionKeys,
        root_key: [u8; 32],
        chain_key: [u8; 32],
        ratchet_key_secret: [u8; 32],
    ) -> Self;

    pub fn active_sending_state(&self) -> Option<ActiveSendingState>;
}

#[cfg(feature = "hazmat-expose-root-key")]
pub struct ActiveSendingState {
    pub root_key: [u8; 32],
    pub chain_key: [u8; 32],
    pub chain_index: u64,
    pub ratchet_key: [u8; 32],
}
```

**No new cryptographic logic.** The constructor composes existing internals (`Ratchet::new_with_ratchet_key`, `RootKey::new`, `ChainKey::new`, `Curve25519SecretKey::from_slice`), and the extractor reads existing fields. Supporting changes are five field-visibility flips from private to `pub(super)` (strictly inside the `session` module) so the cross-module helpers in `double_ratchet.rs` can read the bytes without going through extra accessor wrappers.

## Properties

- **Zero impact when the feature is off.** Default `cargo build` produces identical artifacts to vanilla 0.10.0.
- **No new `unsafe`**, no new dependencies, no new derivation/KDF/AEAD calls.
- **Strict upstream lints respected:** `missing_docs`, `unsafe_code`, `unwrap_used`, `expect_used`, `panic` all pass with `-D warnings`.
- **Existing test suite passes** with the feature enabled (137 lib + 10 doc tests).

## Reference implementation

- Branch: https://github.com/smalyu/vodozemac/tree/hazmat-expose-root-key
- Diff vs 0.10.0: https://github.com/matrix-org/vodozemac/compare/0.10.0...smalyu:vodozemac:hazmat-expose-root-key
- Commit: `5c560684f9400cdca8624c5f70863352a8a31bbe`

The patch is 96 insertions / 5 deletions across 5 files (`Cargo.toml`, `src/olm/session/chain_key.rs`, `src/olm/session/double_ratchet.rs`, `src/olm/session/mod.rs`, `src/olm/session/ratchet.rs`).

## Naming under hazmat convention

The `hazmat-*` prefix follows the convention used by RustCrypto crates, signalling to downstream that callers have already taken responsibility for deriving the root key by a sound mechanism (e.g., a fully analyzed Noise handshake with mutual static authentication).

## My specific use case

Two-party peer-to-peer messenger over Tor v3 onion services. Each peer runs `arti` for inbound onion service + outbound dial; handshake is Noise `KK` over an in-person-exchanged static identity (QR code). I want vodozemac's audited Double Ratchet on top, rather than vendor an unaudited Rust DR implementation.

## Questions for the maintainers

1. Is the `hazmat-*` framing acceptable, or do you prefer a different gating (a separate hazmat module, conditional `pub use`, etc.)?
2. Should the constructor and the equivalence-test extractor live behind separate features (`hazmat-from-root-key` + `hazmat-extract-state`)?
3. If the framing works, I can prepare the PR — happy to adjust naming, granularity, or split.

Thank you for considering.

smalyu:
Closes #334.

Implements the proposal from #334: a feature-gated `Session::from_root_key_material` constructor plus a companion `Session::active_sending_state` accessor, enabling downstream projects to reuse vodozemac's audited Double Ratchet on top of non-Olm handshakes (Noise variants, alternative PQ KEMs, experimental transports).

## Summary

- New `Session::from_root_key_material(config, session_keys, root_key, chain_key, ratchet_key_secret)` constructor — composes existing internals (`Ratchet::new_with_ratchet_key`, `RootKey::new`, `ChainKey::new`, `Curve25519SecretKey::from_slice`); **no new cryptographic logic**.
- New `Session::active_sending_state()` returning a public `ActiveSendingState { root_key, chain_key, chain_index, ratchet_key }` — strictly for equivalence-testing against `Account`-derived sessions.
- Five field-visibility flips from private → `pub(super)` inside the `session` module, so cross-module reads in `double_ratchet.rs` don't need extra accessor wrappers.
- 96 insertions / 5 deletions across 5 files (`Cargo.toml`, `src/olm/session/chain_key.rs`, `src/olm/session/double_ratchet.rs`, `src/olm/session/mod.rs`, `src/olm/session/ratchet.rs`).

## Properties

- **Zero impact when the feature is off** — default `cargo build` produces identical artifacts to vanilla 0.10.0.
- No new `unsafe`, no new dependencies, no new derivation / KDF / AEAD calls.
- Existing test suite passes with the feature enabled (137 lib + 10 doc tests).
- `missing_docs`, `unsafe_code`, `unwrap_used`, `expect_used`, `panic` all pass with `-D warnings`.

## Adjustments per review feedback

Per @poljar's response on the issue (https://github.com/matrix-org/vodozemac/issues/334#issuecomment-4405541573):

- Will move the gate from a new `hazmat-expose-root-key` feature to the existing `low-level-api` feature — happy to push that as a follow-up commit on this branch.
- Keeping the constructor and the extractor under a single feature (not splitting into `hazmat-from-root-key` / `hazmat-extract-state`).

Naming, granularity, or any other details are open to further adjustment during review.

## Use case

Two-party peer-to-peer messenger over Tor v3 onion services. Each peer runs `arti` for inbound onion service + outbound dial; handshake is Noise `KK` over an in-person-exchanged static identity (QR code). The goal is to put vodozemac's audited Double Ratchet on top, rather than vendor an unaudited Rust DR implementation.

### DIFF
--- a/src/olm/session/mod.rs
+++ b/src/olm/session/mod.rs
@@ -884,4 +884,38 @@ mod test {
         let message = message_key.encrypt(plaintext.as_bytes());
         assert_ne!(message.ciphertext, plaintext.as_bytes());
     }
+
+    #[test]
+    #[cfg(feature = "low-level-api")]
+    fn from_root_key_material_exposes_active_sending_state() {
+        use crate::{Curve25519SecretKey, olm::SessionKeys};
+
+        let session_keys = SessionKeys {
+            identity_key: Curve25519PublicKey::from_bytes([0xAA; 32]),
+            base_key: Curve25519PublicKey::from_bytes([0xBB; 32]),
+            one_time_key: Curve25519PublicKey::from_bytes([0xCC; 32]),
+        };
+
+        let root_key = [0x11u8; 32];
+        let chain_key = [0x22u8; 32];
+        let ratchet_key_secret = [0x33u8; 32];
+
+        let session = Session::from_root_key_material(
+            SessionConfig::version_2(),
+            session_keys,
+            root_key,
+            chain_key,
+            ratchet_key_secret,
+        );
+
+        let state = session
+            .active_sending_state()
+            .expect("a freshly constructed Session must have an active sending chain");
+
+        assert_eq!(state.root_key, root_key);
+        assert_eq!(state.chain_key, chain_key);
+        assert_eq!(state.chain_index, 0);
+        let expected_ratchet_key = *Curve25519SecretKey::from_slice(&ratchet_key_secret).to_bytes();
+        assert_eq!(state.ratchet_key, expected_ratchet_key);
+    }
 }
--- a/src/olm/session/mod.rs
+++ b/src/olm/session/mod.rs
@@ -32,6 +32,8 @@ use receiver_chain::ReceiverChain;
 use root_key::RemoteRootKey;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+#[cfg(feature = "low-level-api")]
+use zeroize::Zeroize;
 
 use super::{
     SessionConfig,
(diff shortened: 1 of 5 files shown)


smalyu:
Pushed [205f92e](https://github.com/smalyu/vodozemac/commit/205f92e) — re-gates the new APIs under the existing `low-level-api` feature and drops the dedicated `hazmat-expose-root-key` feature, per @poljar's request on the issue. No semantic change; both `cargo test` (default) and `cargo test --features low-level-api,experimental-session-config` are green.

smalyu:
Pushed two follow-ups addressing the red CI on 205f92e:

- [672153d](https://github.com/smalyu/vodozemac/commit/672153d) — `style: fix import order in double_ratchet.rs`. The cfg-gated `super::ratchet::RatchetKey` import was below the unified `super::{...}` block; nightly rustfmt wanted it above. With this, the CI matrix (clippy + per-feature compile + cross-platform tests) can actually run — they were all `skipped` on the previous run because they depend on `Check style`.

- [e9ed6a7](https://github.com/smalyu/vodozemac/commit/e9ed6a7) — `test(hazmat): cover active_sending_state round-trip`. Adds a `mod test` unit test that builds a `Session` via `from_root_key_material(...)` with deterministic inputs and asserts each field of the returned `ActiveSendingState`. Catches the cargo-mutants survivor on `DoubleRatchet::active_sending_state` (replace body with `None`), which previously had no test coverage.

Local verification (nightly + stable):
- `cargo +nightly fmt --all -- --check` — clean
- `cargo +nightly clippy --all-targets --features low-level-api,experimental-session-config -- -D warnings` — clean
- `cargo test` (default features) — 137 lib + 10 doc pass
- `cargo test --features low-level-api,experimental-session-config` — 147 lib + 10 doc pass, including the new test

Workflows on the new SHA are pending the usual "approve and run for fork PR" gate.

poljar:
Can you please rebase this PR on top of main to get the CI happy?

smalyu:
Rebased on top of `main` (was 19 commits behind). Force-pushed; new tip is [a45b060](https://github.com/smalyu/vodozemac/commit/a45b060). No conflicts — `main` only moved on CI/deps/README. Local verification (`cargo +nightly fmt --check`, `cargo +nightly clippy --all-targets --features low-level-api,experimental-session-config -- -D warnings`, `cargo test` default, `cargo test --features low-level-api,experimental-session-config`) all clean. Thanks @poljar — workflows pending the fork-PR approval.

src/olm/session/mod.rs
```diff
@@ -190,6 +190,51 @@ impl Debug for Session {
...
+#[cfg(feature = "low-level-api")]
+pub struct ActiveSendingState {
+    /// Current root key `R_i` of the active sender chain.
+    pub root_key: [u8; 32],
+    /// Current chain key `C_{i,j}` of the active sender chain.
+    pub chain_key: [u8; 32],
+    /// Index `j` within the current chain.
+    pub chain_index: u64,
+    /// Secret part of our local ratchet key `T_i`.
+    pub ratchet_key: [u8; 32],
```
poljar: These should be boxed as to avoid moves becoming `memcpy` calls which copy secret key material, instead they would copy the pointer if we box this[^1].

[^1]: https://benma.github.io/2020/10/16/rust-zeroize-move.html

src/olm/session/mod.rs
```diff
@@ -190,6 +190,51 @@ impl Debug for Session {
...
+}
+
+#[cfg(feature = "low-level-api")]
+impl Session {
+    /// hazmat: build a `Session` directly from raw active-sender key material
+    /// derived by a non-Olm handshake. No new derivation is performed —
+    /// supplied bytes become the active chain's keys verbatim. The supplied
+    /// [`SessionKeys`] is used solely for the session ID and pre-key message
+    /// envelope, with no further use of its contents.
+    pub fn from_root_key_material(
```
poljar, line 301: Could you expand this comment a bit, let's add:

1. A typical use-case for this
2. An example for how this should be used
3. Explicitly call out that this skips the 3DH step of Olm but the double ratchet part continues on working as usual (this has been called out in the DoubleRatchet method, but this type and the method are private).

poljar (changes requested):
This looks mostly good, it is after all just a new constructor.

Though the docs could be a bit expanded on. There's also an issue with secret key material being copied accidentally.

src/olm/session/mod.rs
smalyu: Done in 20384f1. `root_key`, `chain_key` and `ratchet_key` on `ActiveSendingState` are now `Box<[u8; 32]>` and zeroized on drop, matching the existing `MessageKey`/`RemoteMessageKey` convention. `DoubleRatchet::active_sending_state` now clones the existing boxes (and reuses `Curve25519SecretKey::to_bytes()`) instead of round-tripping the secrets through stack arrays.

src/olm/session/mod.rs
smalyu: Done in 20384f1. Expanded the docs on `from_root_key_material` (and the `ActiveSendingState` type) with: a typical use-case (non-Olm handshakes such as a Noise KK exchange over Tor v3 onion services, alternative/PQ KEMs), a compiling example, and an explicit note that only the Olm 3DH handshake is replaced while the Double Ratchet keeps working as usual.

poljar (approved):

PR #289, opened 2026-01-15
feat: Add support for HPKE

poljar:
This PR implements the bidirectional communication channel described in https://github.com/matrix-org/matrix-spec-proposals/pull/4388.

It depends on #206 as the HPKE crate of our choosing already depends on the bumped version of the Rand crate.

PR #206, opened 2025-01-30
chore: Bump rand to 0.10.0

poljar:
This requires that our dependencies bump to `rand` 0.10.0 first, this includes the Dalek crates and RustCrypto based ones.

This is now mostly ready, we're just waiting for our deps to do a release. We're currently depending on release candidates.

PR #377, opened 2026-07-08
feat: Add support for HPKE

poljar:
This replaces #289.

Cleans up the history a bit and rebases on top of main now that #206 is merged, otherwise it's the same as #289.

I left the branch for #289 alive as the Rust SDK PR uses the git revisions and I didn't want to break the build as I


src/hpke/recipient.rs
```diff
@@ -0,0 +1,258 @@
...
+        message: &InitialMessage,
+        aad: &[u8],
+    ) -> Result<RecipientCreationResult, Error> {
+        let Self { secret_key, application_info_prefix } = self;
+        let InitialMessage { encapsulated_key, ciphertext } = message;
+
+        let their_public_key = *encapsulated_key;
+        let our_public_key = Curve25519PublicKey::from(&secret_key);
+
+        let secret_key = convert_secret_key(&secret_key);
```
dkasak: Since the secret key is never exposed, couldn't we use `hpke::PrivateKey` instead and only do the conversion for the public key?

dkasak (changes requested):
Almost there. I fixed a bunch of things myself (pair reviewing with poljar) and requested a single change. Once that's done, we're good to approve.

src/hpke/recipient.rs
poljar: Did so in https://github.com/matrix-org/vodozemac/pull/377/commits/7f436c09af7662218bf6048bcb06128d0d2e433f.

dkasak (approved):
LGTM

poljar:
Rebased to get the CI going.

PR #379, opened 2026-07-13
Add additive `*_with_rng` entropy-injection variants to the Olm API

shakhvit:
### Summary
Adds additive `*_with_rng` variants to every key-generation entry point in the `olm`
module, letting callers supply their own `impl CryptoRng` instead of the thread-local
`rand::rng()`. Fully backwards-compatible: no existing method changes behaviour.

### Motivation
vodozemac currently sources all key-generation entropy from `rand::rng()` (the
thread-local, OS-seeded generator), with no way to inject a specific RNG. Exposing an
RNG parameter is a standard, widely-useful capability that `x25519-dalek` and
`ed25519-dalek` already provide at the primitive level
(`StaticSecret::random_from_rng`, `SigningKey::generate`). Surfacing it at the
`Account`/`Session` level enables:

- **Deterministic tests / KAT vectors** — reproduce exact keys and ciphertext from a
  fixed seed, so regressions in key-derivation plumbing are caught byte-for-byte.
- **Reproducible builds** and record/replay harnesses.
- **Custom or hardware entropy sources** — HSMs/TPMs, or a host that already owns a
  vetted CSPRNG and wants a single entropy source.
- **`no_std`/embedded targets** where `getrandom` has no backend and entropy must be
  provided by the caller.

### What changed
New `*_with_rng` methods, each taking `rng: &mut impl CryptoRng`:
- `Account::new_with_rng`
- `Account::generate_one_time_keys_with_rng`
- `Account::generate_fallback_key_with_rng`
- `Account::create_outbound_session_with_rng`
- `Session::encrypt_with_rng` — consumes RNG bytes **only** when the message triggers a
  Diffie-Hellman ratchet advance; a pure symmetric-chain message draws nothing.
- `Curve25519SecretKey::new_with_rng`, `Curve25519Keypair::new_with_rng`
- `Ed25519Keypair::new_with_rng`, `Ed25519SecretKey::new_with_rng`

The rng is threaded through internal helpers (`Ratchet`, `RemoteRootKey::advance`,
`DoubleRatchet`, `OneTimeKeys`, `FallbackKeys`) via parallel `_with_rng` methods.
`decrypt`/`create_inbound_session` are unchanged — they mint no keys.

### Non-breaking / additive
Every existing method keeps its exact body and behaviour, still backed by `rand::rng()`.
The diff is purely additive (the only two removed lines augment two `use` statements). No
public signature changes; no existing test changed. The audited `OsRng` code paths are
byte-identical.

### Security note (the footgun)
An injected RNG is a sharp tool: the security of every key rests entirely on the quality
of the supplied generator. A low-entropy, predictable, or **reused** RNG produces
predictable or repeated secret keys, and reusing an RNG state across two genuinely-new
Diffie-Hellman ratchet steps collapses the ephemeral keys and destroys forward secrecy /
post-compromise security. Each new public method documents this explicitly and directs
callers to pass a cryptographically secure generator seeded with fresh entropy per
genuinely-new operation. The default `OsRng`-backed methods remain the right choice for
almost all users.

### Tests
`tests/with_rng.rs` covers: determinism (identical RNG ⇒ byte-identical keys/session-id/
ciphertext across all seams); distinct RNG ⇒ distinct keys; interop (a `_with_rng`-built
session round-trips with a default `OsRng`-built peer); and the lazy DH-ratchet advance
seam (reproducible under the same RNG, fresh ephemeral under a different one). All
existing tests pass; `cargo clippy --all-targets --all-features` is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Add additive `*_with_rng` variants to every `olm` key-generation entry point
+  (`Account::new`, `Account::generate_one_time_keys`,
+  `Account::generate_fallback_key`, `Account::create_outbound_session` and
+  `Session::encrypt`, plus the `Curve25519SecretKey`/`Ed25519Keypair`/
+  `Ed25519SecretKey` constructors). These accept a caller-supplied
+  `impl CryptoRng` so randomness can be injected for deterministic testing,
+  reproducible builds and custom/hardware entropy sources. The existing
+  `OsRng`-backed methods are unchanged.
 - Add a new default-enabled `precomputed-tables` feature flag which controls
   curve25519-dalek's precomputed basepoint tables. Size-sensitive builds can
   now disable the default features to drop roughly 40 KB of lookup tables from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
   `impl CryptoRng` so randomness can be injected for deterministic testing,
   reproducible builds and custom/hardware entropy sources. The existing
   `OsRng`-backed methods are unchanged.
+  ([#379](https://github.com/matrix-org/vodozemac/pull/379))
 - Add a new default-enabled `precomputed-tables` feature flag which controls
   curve25519-dalek's precomputed basepoint tables. Size-sensitive builds can
   now disable the default features to drop roughly 40 KB of lookup tables from
--- a/src/olm/account/fallback_keys.rs
+++ b/src/olm/account/fallback_keys.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use rand::CryptoRng;
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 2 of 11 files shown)


poljar (changes requested):
Is this something you are personally invested in? What's your use-case here?

This PR needs some more work:

1. The `*_with_rng()` functions should be called by the current methods with the RNG set to `UnwrapErr(SysRng)`.
2. The PR should be split out into smaller commits, ideally one added method would be one commit.
3. If this is done for the `olm` submodule we should do this for the other submodules as well, could be a separate PR.

shakhvit:
@poljar  thanks for the response.

Yes, I wanted to pass my own entropy into your primitives to have full control over the entropy in the system and drop the OS-random dependency. That's actually what let me compile Olm for a bare-metal target (no `getrandom` backend), which is the main thing I'm after. I'm maintaining a downstream project built on vodozemac, so this isn't a one-off. I'm invested in getting it right and keeping it maintained upstream. Give me a few days to push the reworked version. Thanks for taking the time to review!

MatMaul:
@shakhvit do you plan to get back at it soon ?

We would need pretty quickly on our side, and I am happy to take over if you don't have the bandwidth for it.

shakhvit:
@MatMaul I had no time for this recently. Yes, please, you can take it over. Thanks :)

poljar:
#409 has improved on this PR, so I'll close this one.

Issue #382, opened 2026-07-14
`from_libolm_pickle` mishandles the one-time-key counter, invalidating the newest OTK on the next `generate_one_time_keys`


alinradut:
We're porting an app off libolm and onto vodozemac, and while testing the account import path I ran into a bug that quietly breaks one-time keys. Nothing goes wrong at import time; the damage only shows up later, on the first key replenishment, and even then it's silent until some peer tries to talk to you.

Short version: `Account::from_libolm_pickle` copies libolm's one-time-key id counter across as-is, but the two implementations mean different things by it. libolm stores the *last id it handed out*; vodozemac reads it as the *next id to hand out*. So they're off by one, and the first `generate_one_time_keys()` after an import reuses the id of the most recently imported key and overwrites its private half.

If that key had already been published (which it will have been, in any real account), you now have a published one-time key whose private key is gone. When a peer claims it and sends you a pre-key message, session setup fails. It's a permanent UTD for that session until something forces a new one.

Seen on vodozemac 0.10.0. Source pickles from libolm 3.2.16.

I also put together a self-contained repro  ([main.rs.txt](https://github.com/user-attachments/files/30008850/main.rs.txt)) that goes all the way to the session failure (and shows it working once the counter is fixed), depending only on vodozemac. It embeds a real libolm 3.2.16 pickle so there's no libolm build involved.

```rust
// libolm side: 5 one-time keys, published, no fallback key.
let olm = OlmAccount::new();
olm.generate_one_time_keys(5);
olm.mark_keys_as_published();
let pickle = olm.pickle(PicklingMode::Encrypted { key: KEY.to_vec() });

// vodozemac side.
let mut acc = Account::from_libolm_pickle(&pickle, KEY).unwrap();
assert_eq!(acc.stored_one_time_key_count(), 5);

acc.generate_one_time_keys(1);
assert_eq!(acc.stored_one_time_key_count(), 6); // fails: still 5, one got overwritten
```

## Root cause

libolm hands out one-time-key ids with a pre-increment and then pickles the counter as the last value it used (`src/account.cpp`):

```cpp
key.id = ++next_one_time_key_id;                    // ~287
current_fallback_key.id = ++next_one_time_key_id;   // ~310, fallback shares the counter
pos = olm::pickle(pos, value.next_one_time_key_id); // ~517, pickled as-is
```

So five OTKs and no fallback key gives you `next_one_time_key_id == 5`.

[...]

PR #383, opened 2026-07-15
fix: libolm unpickling invalidates last otk

alinradut:
This fixes issue #382 where Account::from_libolm_pickle would overwrite the last olm one-time-key the next time a one-time-key would be generated.


poljar (changes requested):
Thanks for the patch and the find.

Looks good, just needs a changelog entry.

Please add one to the top of the current entries in the unreleased section.

poljar (approved):
Thanks again.

PR #384, opened 2026-07-20
chore: Enable the unreachable_pub Rust lint

poljar:
This makes it more explicit and obvious which items are public and which are not.

It also caught some re-exports that didn't make it to the public API level. Some exceptions are also necessary due to us having exports available only under a feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ unsafe_op_in_unsafe_fn = "deny"
 unused_extern_crates = "deny"
 unused_import_braces = "deny"
 unused_qualifications = "deny"
+unreachable_pub = "deny"
 rust_2018_idioms = "deny"
 
 [lints.clippy]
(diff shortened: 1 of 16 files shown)


richvdh (approved):
> Some exceptions are also necessary due to us having exports available only under a feature flag.

I assume we can't fix this by, say, setting all the feature flags when running the lint checks?

poljar:
> > Some exceptions are also necessary due to us having exports available only under a feature flag.
> 
> I assume we can't fix this by, say, setting all the feature flags when running the lint checks?

We can do that for some of them, but these lints are commonly shown in editors as well. So you'll have to change your editor config or enable the feature flag by default.

There are also some functions which replace our normal functions during fuzzing (the `fuzzing` cfg flag is enabled). If we're in fuzzing mode, a different set of functions becomes unreachable.

PR #403, opened 2026-08-19
Add feature to directly use the OS-provided RNG instead of `rand`

MatMaul:
The os-rng feature enables the use of the operating system's RNG instead of the default random number generator provided by the `rand` crate, which is not backtracking resistant.

This is usually not necessary, but can be enabled for security-critical applications or compliance purposes.

Please note that OS RNG is not guaranteed to be backtracking resistant either since it will depend on the underlying OS implementation, but the RNG state will be available in kernel space instead of user space so harder to extract.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,11 @@ experimental-session-config = []
 # The low-level-api feature exposes extra APIs that are only useful in advanced
 # use cases and require extra care to use.
 low-level-api = []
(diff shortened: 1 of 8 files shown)


poljar:
Isn't this more generally solved by https://github.com/matrix-org/vodozemac/pull/379?

MatMaul:
Indeed thanks!

PR #404, opened 2026-08-25
Ignore a mutation that would cause an infinite loop

poljar:


poljar:
We don't need this if we merge #405 as we're going to just remove the function we're testing.

poljar:
As mentioned, #405 made this redundant.

PR #405, opened 2026-08-25
Mark some of the indices in olm/megolm messages as required

poljar:
This ensures that the fields won't be omitted if they contain the default value 0.

This removes the need for our custom encode_manual methods while retaining libolm compatibility.

Took a while before coming back to this.

### DIFF
--- a/src/megolm/message.rs
+++ b/src/megolm/message.rs
@@ -22,7 +22,7 @@ use crate::{
     DecodeError,
     cipher::{Cipher, Mac, MessageMac},
     types::{Ed25519Keypair, Ed25519Signature},
-    utilities::{VarInt, base64_decode, base64_encode, extract_mac},
+    utilities::{base64_decode, base64_encode, encode_protobuf_message, extract_mac},
 };
 #[cfg(feature = "low-level-api")]
 use crate::{Ed25519PublicKey, SignatureError};
@@ -143,7 +143,7 @@ impl MegolmMessage {
             ciphertext: self.ciphertext.clone(),
         };
 
-        message.encode_manual(self.version)
+        encode_protobuf_message(message, self.version)
     }
 
     fn set_mac(&mut self, mac: Mac) {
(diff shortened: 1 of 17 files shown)


src/megolm/message.rs
```diff
@@ -143,7 +143,15 @@ impl MegolmMessage {
             ciphertext: self.ciphertext.clone(),
         };
 
-        message.encode_manual(self.version)
+        let mut output: Vec<u8> = vec![0u8; message.encoded_len() + 1];
+        output[0] = self.version;
+
+        #[allow(clippy::expect_used)]
```
andybalaam: This `allow` is not needed, presumably because the `expect` does the job.

src/olm/messages/message.rs
```diff
@@ -143,12 +143,21 @@ impl Message {
...
-        }
-        .encode_manual(self.version)
+        };
+
+        let mut output: Vec<u8> = vec![0u8; message.encoded_len() + 1];
+        output[0] = self.version();
+
+        #[allow(clippy::expect_used)]
+        message
+            .encode(&mut output[1..].as_mut())
```
andybalaam: Can we avoid a second copy of this code somehow?

andybalaam (approved):
Awesome to delete this cod! Couple of small changes. Looks good to me when they are addressed.

src/megolm/message.rs
poljar: What do you mean by "the expect does the job"?

Vodozemac is a bit more restrictive compared to the SDK repo, `expect()` [calls are disallowed throughout the codebase](https://github.com/matrix-org/vodozemac/blob/30d27bb5603fdc3e7880fe7477deeb7f0245eef8/Cargo.toml#L36).

Now sometimes we still need to call an expect here and there, and thus we need to add case-by-case exceptions.

You might have misunderstood this for a unwrap call being allowed here?

src/olm/messages/message.rs
poljar: image

https://github.com/matrix-org/vodozemac/pull/405/commits/07fb29a4bfba3461876c70af2f0bd05e7a08e2ec

poljar:
Oh no, the `cargo fmt` changes have arrived here as well.

src/megolm/message.rs
andybalaam: Ah sorry yeah I totally misread this as a `must_use` allow, sorry.

src/olm/messages/message.rs
andybalaam: \o/

andybalaam (approved):

PR #409, opened 2026-09-01
Add additive `*_with_rng` entropy-injection variants to the API

MatMaul:
PR addressing comments on #379.

Thanks @shakhvit for the initial work.

It also adds a `disallow-default-rng` feature to disable public methods that use the default RNG. This will force the caller to provide a custom RNG via `with_rng` methods instead.


MatMaul:
I tried to use that in the `matrix-rust-sdk`, and we would need to change a bunch of calls there to always use the `with_rng` versions.

@poljar in your review [here](https://github.com/matrix-org/vodozemac/pull/379#pullrequestreview-4713110810) you mention:
`The *_with_rng() functions should be called by the current methods with the RNG set to UnwrapErr(SysRng)`

That is not the current behavior, since randomness is draw from a thread local CSPNRG initialized with `SysRng`, and not `SysRng` directly.
If we want to change that to use `SysRng` directly by default I am all for that, since that is what I am asked to do for compliance.

MatMaul:
Here are the matching needed changes in `matrix-rust-sdk`.
https://github.com/tchapgouv/matrix-rust-sdk/commit/3ab38030593121d3fe4f022eae213af581d46392

I haven't changed the tests yet, let's see where we want to go regarding what my open questions in previous comment.

poljar:
> That is not the current behavior, since randomness is draw from a thread local CSPNRG initialized with `SysRng`, and not `SysRng` directly. If we want to change that to use `SysRng` directly by default I am all for that, since that is what I am asked to do for compliance.

Yeah I'm aware. This would align us with the rest of the RustCrypto ecosystem so that's why I suggested that in the previous PR as well.

If that helps you and you can omit having Rust SDK changes, then that's great.

src/ecies/mod.rs
```diff
@@ -94,7 +94,8 @@ pub use crate::hpke::{CheckCode, DigitMode};
 
 mod messages;
 
-const MATRIX_QR_LOGIN_INFO_PREFIX: &str = "MATRIX_QR_CODE_LOGIN";
+/// The prefix used for Matrix QR login info in ECIES messages.
+pub const MATRIX_QR_LOGIN_INFO_PREFIX: &str = "MATRIX_QR_CODE_LOGIN";
```
poljar, line 98: Why is that pub now?

src/ecies/mod.rs
```diff
@@ -179,6 +180,7 @@ impl Ecies {
     /// this for a different purpose, consider using the [`Ecies::with_info()`]
     /// method.
     #[allow(clippy::new_without_default)]
+    #[cfg(not(feature = "disallow-default-rng"))]
```
poljar, line 183: This is a bit opposite of how it should be done.

Enabling the feature should enable the availability of the `new()` method.

To align with the wider ecosystem the feature should be called `getrandom` and like previously discussed, it should use [SysRng](https://docs.rs/getrandom/latest/getrandom/struct.SysRng.html).

src/ecies/mod.rs
```diff
@@ -82,7 +82,7 @@
 
 use chacha20poly1305::{ChaCha20Poly1305, Key as Chacha20Key, KeyInit, Nonce, aead::Aead};
 use hkdf::Hkdf;
-use rand::rng;
+use rand::CryptoRng;
```
poljar, line 85: We probably want to stop depending on `rand` and instead get `CryptoRng` from `rand_core`.

tests/with_rng.rs
```diff
@@ -0,0 +1,260 @@
...
+        msg1.to_parts(),
+        msg2.to_parts(),
+        "ciphertext must be byte-identical for identical (state, seed, plaintext)"
+    );
+}
+
+#[test]
+fn with_rng_session_interoperates_with_default_account() {
+    // Bob is built with the default (OsRng) path; Alice uses the `_with_rng`
+    // path. If the seam is behaviour-preserving they must be able to talk.
```
poljar, line 124: Is that some idiom I'm unfamiliar with? What does "If the seam is behavior-preserving" mean in this context? What would be the seam here?

tests/with_rng.rs
```diff
@@ -0,0 +1,260 @@
...
+            bob_otk,
+            &mut seeded(21),
+        )
+        .expect("outbound session");
+
+    let message =
+        alice_session.encrypt_with_rng("hello from with_rng", &mut seeded(22)).expect("encrypt");
+    let OlmMessage::PreKey(prekey) = message else {
+        panic!("first message must be a pre-key message");
+    };
```
poljar, line 142: We have since stated to use `assert_matches` instead of manually panicking for such deconstruction.

tests/with_rng.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    alice_session
+}
+
+#[test]
+fn dh_ratchet_advance_is_reproducible_under_same_rng() {
+    // Same session state + same advance RNG => byte-identical advancing message,
+    // including the freshly minted ratchet public key embedded in the header.
+    //
+    // This is the load-bearing assertion for the encrypt seam: it pins that the
```
poljar, line 195: This is some AI mumbo-jumbo, can we reword this whole thing?

poljar (changes requested):
Yeah, this is more in line with how this kind of feature should look like.

Still some things need to be a bit reworked before we're aligned with the rest of the ecosystem with regards to the `with_rng()` methods.

src/ecies/mod.rs
MatMaul: It was because I only added `with_info_and_rng` at first, I've removed the pub and added `new_with_rng` too instead.

tests/with_rng.rs
MatMaul: LLM nonsense, removed, thanks.

